PR #51, opened 2015-01-21
Support passing kwargs to `requests.request` in `apexecute`

Trii:
This allows you to pass other keyword args into `apexecute` for timeouts, etc


nickcatal:
This looks awesome! Thanks

PR #52, opened 2015-01-23
Fix upsert/update/delete to optionally return response body

fr-anthony:
When using `upsert` to create a record, it's convenient to be able to inspect the response. Added similar code to update and delete for symmetry.

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -454,11 +454,7 @@ class SFType(object):
         """
         result = self._call_salesforce('PATCH', self.base_url + record_id,
                                        data=json.dumps(data))
-        if response_body:
-            return result.status_code
-        else:
-            return (result.status_code,
-                    result.json(object_pairs_hook=OrderedDict))
+        return self._response_body(result, response_body)
 
     def update(self, record_id, data, response_body=False):
         """Updates an SObject using a PATCH to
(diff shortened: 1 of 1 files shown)


nickcatal:
Good idea. This is one design decision that I made early on that has always bugged me.

Can you add details about `response_body=True` to the arguments list in the docstrings?

Thanks!

Trii:
This also changes the return value of a few basic methods to be `(int, dict)` instead of the `int`. That is probably important to tell people as well.

nickcatal:
@Trii right, but not by default. by default it continues to return an `int` unless you add an argument

My other idea was that there perhaps should be a way to just get the raw `requests` response so you can do whatever the hell you want with it. You'd have to have a basic understanding of the `requests` library to do anything useful with it, but it would solve this problem.

Ideas?

Trii:
I like the idea of using the flag to return a tuple of `(int, requests.Response)` for sure!

nickcatal:
@Trii in that case should we even return the `int`? You could easily get the status code back from the `Response` object.

Trii:
I agree that just the response would suffice. Using the new flag to control either `int` or `Response` would be a good way to not break BC but give the methods a more flexible behavior.

nickcatal:
@fr-anthony Awesome! Thanks for updating the PR to reflect the change! This will be really useful

PR #54, opened 2015-01-27
simple wrapper for tooling api requests

vnehess:
simple copy of the restful() method with the tooling API path added for connivence.

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -121,6 +121,9 @@ class Salesforce(object):
                                  version=self.sf_version))
         self.apex_url = ('https://{instance}/services/apexrest/'
                          .format(instance=self.sf_instance))
+        self.tool_url = ('https://{instance}/services/data/v{version}/tooling/'
+                            .format(instance=self.sf_instance,
+                                 version=self.sf_version))
 
     def describe(self):
         url = self.base_url + "sobjects"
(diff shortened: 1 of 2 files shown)


Trii:
This is pretty useful. Could you add `**kwargs` to `tooling()` and pass it through to `self.request.get()` the same way the `_call_salesforce()` methods do? The `restful()` method, as well as a few others, are missing it.

vnehess:
Ok 
Sure

simple_salesforce/api.py
```diff
@@ -202,6 +205,27 @@ def restful(self,path,params):
             return None
         else:
             return json_result
+    
+    # Generic Tooling API Function
+    def tooling(self,path,params):
```
nickcatal: Can you add spaces between the arguments here?

simple_salesforce/api.py
```diff
@@ -202,6 +205,27 @@ def restful(self,path,params):
...
+
+        * path: The path of the request
+            Example: sobjects/User'
+        * params: dict of parameters to pass to the path
+        """
+
+        url = self.tool_url + path
+        result = self.request.get(url, headers=self.headers, params=params)
+        if result.status_code != 200:
+            raise SalesforceGeneralError(result.content)
```
nickcatal: Can you do `_exception_handler(result)` (see bottom of `api.py`) instead of throwing `SalesforceGeneralError()` directly?

I now see that there are other parts of the file that don't use that (and I'll have to fix them) but it's the only way for folks to correctly capture the correct exception for the result.

Thanks

demianbrecht:
This is super old, but would still be nice to get in if tests were added.

nullglitch:
can anyone with write access merge this branch with the master? The tooling function is a must have which is missing in the master.

demianbrecht:
@nullglitch why is this function a must have? It's a convenience version of `restful`.

tanderegg:
@nullglitch It looks like there are a few outstanding issues with this PR, if someone wants to address them then I can merge the PR.  See @nickcatal's comments under "Files Changed", plus @bolaurent's error report.

amd-yan:
> @nullglitch why is this function a must have? It's a convenience version of `restful`.

I had the same same question until I realized that there are certain things can only be updated using tooling API today. Eg: Global Picklist.

amd-yan:
I spent sometime this afternoon and added tooling API to the latest code. It works to get my job done. I also created a pull request, and hopefully, other people may find it useful too.

jon-wobken:
Resolved by #445

PR #55, opened 2015-02-02
Retain useful information and store in Exception object

### LINKED ISSUE #10: Pass error fields to exception
Currently, API errors are handed off to Exceptions, but only after being formatted into the `message` field as a string, which forces folks to use regex to get at them.

Attach these values to the error, rather than locking them away in `message`.

jsok:
Closes #10.

Standardise the Exception data and make the exception handler a little more elegant.

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -598,7 +598,6 @@ class SalesforceAPI(Salesforce):
 
 def _exception_handler(result, name=""):
     """Exception router. Determines which error to raise for bad results"""
-    url = result.url
     try:
         response_content = result.json()
     except Exception:
@@ -613,7 +612,7 @@ def _exception_handler(result, name=""):
     }
     exc_cls = exc_map.get(result.status_code, SalesforceGeneralError)
 
-    raise exc_cls(url, result.status_code, name, response_content)
+    raise exc_cls(result.url, result.status_code, name, response_content)
 
 
 class SalesforceMoreThanOneRecord(SalesforceError):
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -100,7 +100,7 @@ class Salesforce(object):
                 proxies=self.proxies)
 
         else:
-            raise SalesforceGeneralError(
+            raise TypeError(
                 'You must provide login information or an instance and token'
             )
 
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -101,7 +101,8 @@ class Salesforce(object):
 
         else:
             raise SalesforceGeneralError(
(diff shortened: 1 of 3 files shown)


simple_salesforce/util.py
```diff
@@ -34,4 +34,14 @@ def date_to_iso8601(date):
...
+
+    message = 'Unknown error occurred for {url}. Response content: {content}'
+
+    def __init__(self, url, status, resource_name, content):
+        self.url = url
+        self.status = status
+        self.resource_name = resource_name
+        self.content = content
+
+    def __str__(self):
```
nickcatal: This should probably be `__unicode__(self):` right?

You'll probably want to make sure to do `message = u''` above as well

simple_salesforce/api.py
```diff
@@ -586,30 +604,16 @@ def _exception_handler(result, name=""):
...
+    exc_map = {
+        300: SalesforceMoreThanOneRecord,
+        400: SalesforceMalformedRequest,
+        401: SalesforceExpiredSession,
+        403: SalesforceRefusedRequest,
+        404: SalesforceResourceNotFound,
+    }
+    exc_cls = exc_map.get(result.status_code, SalesforceGeneralError)
+
+    raise exc_cls(url, result.status_code, name, response_content)
```
nickcatal: Smart!

simple_salesforce/util.py
jsok: Yeah I've never been sure how Python 3 did exception message stringification. Closest thing I could find was https://docs.python.org/3.4/tutorial/errors.html#user-defined-exceptions but it's using `__str__` too!

simple_salesforce/api.py
nickcatal: You could probably also pass in `result.url` as the first argument and remove the `url = result.url` higher up.

simple_salesforce/util.py
nickcatal: Hmm...

We could require `six` to be installed or if we don't want to require another library add [the code behind](https://bitbucket.org/gutworth/six/src/78315705481c3f5e491a3816e41aa66c54c8f965/six.py?at=default#cl-803) `python_2_unicode_compatible` to the utils file.

jsok: Does my latest commit address this issue? Or is it insufficient?

nickcatal: Hmm... technically `__str__()` should not return `unicode` on Python2, but I'm failing to see where that'd be an issue in this case. So... sure! Thanks for the PR

jsok:
Thanks! Can you please update PyPi as well :)

nickcatal:
Done! 0.67.0 is live

Issue #60, opened 2015-04-26
Unable to serialize object Salesforce


laubzega:
I need to serialize object Salesforce so that I can store it in Django's request.session for repeated use. However, neither JSON- nor Pickle-based serializers can handle it. I ended up adding following code to `Salesforce.__getattr__()`
```python
        # fix that enables pickle to work
        if name.startswith('__') and name.endswith('__'):
            return super(Salesforce, self).__getattr__(name)
```
to eliminate problems caused by special attributes, but there may be a better solution.

nickcatal:
I really like your fix. Considering Salesforce objects will never start with 2 leading underscores I think changing the library so that mangled attributes are always returned with `super(Salesforce, self).__getattr__(name)` makes perfect sense.

If you can, please do a pull request that has something along the lines of

```python
    if name.startswith('__'):
        return super(Salesforce, self).__getattr__(name)
```

demianbrecht:
Closing as fix was merged.

PR #62, opened 2015-05-02
Fix for Salesforce object serialization

laubzega:
Here is proposed fix for #60.


nickcatal:
This looks good! Thanks!

@scottpersinger can you merge this in

PR #64, opened 2015-05-15
Fix typos

bennylope:
Features are deprecated, not depreciated (unless you're an accountant!).


demianbrecht:
Thanks for the catch!

PR #69, opened 2015-07-16
Add queryAll to retrieve more complete set of records

SoftwareArtisan:
add keyword query_all = True to query, query_more, query_all in support of /queryAll SF REST request

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -267,7 +267,7 @@ class Salesforce(object):
         * query -- the SOQL query to send to Salesforce, e.g.
                    `SELECT Id FROM Lead WHERE Email = "waldo@somewhere.com"`
         """
-        url = self.base_url + 'query/'
+        url = self.base_url + ( 'queryAll/' if kwargs.get('queryAll') else 'query/' )
         params = {'q': query}
         # `requests` will correctly encode the query string passed as `params`
         result = self.request.get(url, headers=self.headers, params=params, **kwargs)
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -267,7 +267,7 @@ class Salesforce(object):
         * query -- the SOQL query to send to Salesforce, e.g.
                    `SELECT Id FROM Lead WHERE Email = "waldo@somewhere.com"`
         """
-        url = self.base_url + ( 'queryAll/' if kwargs.get('queryAll') else 'query/' )
+        url = self.base_url + ( 'queryAll/' if ( 'query_all' in kwargs and kwargs.pop('query_all') ) else 'query/' )
         params = {'q': query}
         # `requests` will correctly encode the query string passed as `params`
         result = self.request.get(url, headers=self.headers, params=params, **kwargs)


simple_salesforce/api.py
```diff
@@ -267,7 +267,7 @@ def query(self, query, **kwargs):
         * query -- the SOQL query to send to Salesforce, e.g.
                    `SELECT Id FROM Lead WHERE Email = "waldo@somewhere.com"`
         """
-        url = self.base_url + 'query/'
+        url = self.base_url + ( 'queryAll/' if ( 'query_all' in kwargs and kwargs.pop('query_all') ) else 'query/' )
```
nickcatal: Not a huge fan of having this all inline (plus it goes past the 80 char pep8 limit)

If we decide to go this route we'll probably need refactor this, then add something to the docstring.

nickcatal:
I'm conflicted here. I created the `query_all()` method before that endpoint existed on the API (or before I knew about it, I forget)

Does it make sense to have this as part of `Salesforce.query()` at all?

SoftwareArtisan:
I'd be happy to make it either a standalone method or make the keyword "official" instead of leveraging backdoor kwargs.

As a standalone method I'd still want to leverage query_all capability - returning full range of records.  Again, we could have 3 parallel constructions - queryXXX, queryXXX_all, queryXXX_more.... or, and I think this is cleaner, pass a known flag down through the call chain query(...,all=True).

What was the use case for pushing kwargs to the request method?  I used the one-liner to remove the keyword flag to avoid blowing it up.

SoftwareArtisan:
Would like to get closure on this issue.  Please let me know how you would like the code to look or if this is not going to be integrated at all.  Thanks.

simple_salesforce/api.py
```diff
@@ -266,8 +266,10 @@ def query(self, query, **kwargs):
 
         * query -- the SOQL query to send to Salesforce, e.g.
                    `SELECT Id FROM Lead WHERE Email = "waldo@somewhere.com"`
+
+        * query_all: if True queries for all records, archived/deleted/active else just active set
         """
-        url = self.base_url + 'query/'
+        url = self.base_url + ( 'queryAll/' if ( 'query_all' in kwargs and kwargs.pop('query_all') ) else 'query/' )
```
Trii: You don't need the first `'query_all' in kwargs`. 

```python
url = self.base_url + 'queryAll/' if kwargs.pop('query_all', False) else 'query/'
```

dernetterick:
Hi,
why is this still open? I was about to create a pull request for the same purpose when I stumbled across this one.

demianbrecht:
Apologies for the delay on getting this merged. We're currently working through some legal details around potential ownership change and are unable to merge PRs until they're all sorted.

gregpinero:
Hi guys,  is this change available on pip?

demianbrecht:
Hey @gregpinero. Sorry, not yet as I'm still unable to merge PRs from external sources. Hoping to have this resolved soon!

gregpinero:
Hey @demianbrecht No problem.  Do you have a general idea if it's day or weeks?  If it's more than a few days I'll just use this modification.

demianbrecht:
@gregpinero: unfortunately it seems like it may be a few more days. I'm hoping to have everything tied up in the next week, but can't promise anything.

gregpinero:
Hi guys, I just wanted to check on that status of this.  Is it merged?  Available via pip?

demianbrecht:
Hi @gregpinero, apologies on the delay on this, it's be _far_ more problematic than anticipated to get things sorted. We're on the final stretch here and I'm pushing to have any blockers removed as soon as possible.

For a little background on why this is taking so long, I was working with the Salesforce legal team to attempt to take this on as an official Salesforce project, which is why PRs were initially locked down. Unfortunately there were some blockers on making that happen, so we're now in the process of tying up some loose ends, but the gates should be lifted soon enough.

My apologies and thanks for the understanding!

SoftwareArtisan:
> Hi @gregpinero, apologies on the delay on this, it's be _far_ more
> problematic than anticipated to get things sorted. We're on the final
> stretch here and I'm pushing to have any blockers removed as soon as
> possible.
>
> For a little background on why this is taking so long, I was working with
> the Salesforce legal team to attempt to take this on as an official
> Salesforce project, which is why PRs were initially locked down.
> Unfortunately there were some blockers on making that happen, so we're now
> in the process of tying up some loose ends, but the gates should be lifted
> soon enough.
>
> My apologies and thanks for the understanding!
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/heroku/simple-salesforce/pull/69#issuecomment-185540172

Sounds good.  I'm offline for a few weeks and won't be able to get to it
until April in any case.

Best of luck with Salesforce folks,

Will

demianbrecht:
So, with being good again to push changes, if a couple tests are added (with/without arg) for this and the comment that @Trii left is addressed, this will be good to merge.

demianbrecht:
Okay, so getting my bearings on this again...

`query_all` is to deal with pagination, whereas the `query_all` flag is a flag for the `queryAll` endpoint, which includes deleted records. Anything missing there? If not, unfortunately I think that with the existence of the `query_all` method, the parameter name is confusing and should likely be renamed to `deleted` or something along those lines.

Additionally, with the implementation as-is, the `queryAll` endpoint is only accessible if calling `query` and not `query_all` (this should be changed for consistency).

Also, tests will need to be added for this change.

kvarga:
Thoughts on using just the .query() method to support all of these?

i.e. `sf.query("query string", all_results=True, include_deleted=True)`

This would be backwards compatible, and then just add a deprecation statement to calls to query_all. We could work on nomenclature of `all_results` and `include_deleted` as well

SoftwareArtisan:
Sounds very straightforward. No new API required, backward compatible.

Guess to thrash it out completely we have to realize a pedant might 
argue what "all" means.  Is "all" all records deleted or not?  Or "all" 
active records (which is how we're suggesting with this change)?  In any 
case, this change would, to be selfish about it :-), satisfy my use-case.

kvarga:
salesforce nomenclature uses "results" and "records" interchangeably and queryAll for including deleted items.

I think using `all_results` and `include_deleted` might be descriptive enough?

digitalaun:
Worked on this in #152 prior to seeing the new comments made 6 days ago. Only went with supporting `include_deleted` for now.

PR #72, opened 2015-08-23
Custom session injection

### LINKED ISSUE #17: No support for proxies
I need to go through a proxy server to access salesforce. Could we pass in proxy object to all requests?

### LINKED ISSUE #71: Add the ability to define an existing requests.Session for Salesforce instance
I have a use case where I'm using a flow-based programming framework, where one of the workflow nodes executes a login. Because task nodes are intended to only have serializable data returned in order to support multiprocessing-based parallelization, I can return the session ID. In subsequent tasks that execute that are dependent on a Salesforce login, I can then take that session ID and pass it into a new salesforce instance with the "direct" instantiation. Unfortunately, this causes a unique requests.Session per instance.

What would be nice is if I could define the requests.Session instance in my own code external to the tasks and pass it into Salesforce.__init__ to preserve the session for the duration of execution.

demianbrecht:
Allows for the injection of session objects created outside of
simple_salesforce, exposing some of the more advanced features of
requests and urllib3 to the user.

This fixes #71.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,18 +4,18 @@ from setuptools import setup
 import textwrap
 import sys
 
-extra_install_requires = []
-extra_tests_require = []
+pyver_install_requires = []
+pyver_tests_require = []
 if sys.version_info < (2, 7):
-    extra_install_requires.append('ordereddict>=1.1')
-    extra_tests_require.append('unittest2>=0.5.1')
+    pyver_install_requires.append('ordereddict>=1.1')
+    pyver_tests_require.append('unittest2>=0.5.1')
 
 if sys.version_info < (3, 0):
-    extra_tests_require.append('mock==1.0.1')
+    pyver_tests_require.append('mock==1.0.1')
 
 setup(
     name='simple-salesforce',
-    version='0.67.1',
+    version='0.68.0',
     author='Nick Catalano',
     packages=['simple_salesforce',],
     url='https://github.com/neworganizing/simple-salesforce',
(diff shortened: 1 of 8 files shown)


demianbrecht:
@nickcatal: mind taking a poke at this when you get a chance? If you don't have any objections I'll do a minor version bump and merge. It's a feature that would be useful for a project I'm currently working on.

Cheers!

simple_salesforce/login.py
```diff
@@ -12,7 +12,7 @@
 import requests
 
 
-def SalesforceLogin(**kwargs):
+def SalesforceLogin(session=None, **kwargs):
```
nickcatal: Based on the convention below I've been doing `kwargs.get('session')` instead of defining it inside the actual definition line.

I'm open to switching obviously. But mixing both doesn't make a ton of sense.

Also you'll probably want to add something to the docstring about this.

setup.py
```diff
@@ -5,7 +5,7 @@
 import sys
 
 extra_install_requires = []
-extra_tests_require = []
+extra_tests_require = ['httpretty']
```
nickcatal: Instead of having this here, put it inside `tests_require`
Below `pytz`

These lists are intentionally empty then appended to based on if a specific package is needed for a specific version of python.

I should probably spell this out more with documentation.

simple_salesforce/login.py
demianbrecht: > Based on the convention below I've been doing kwargs.get('session') instead of defining it inside the actual definition line.

Yeah, I'm generally not a fan of that for two reasons: 

1. It doesn't really make sense if keywords aren't arbitrary (i.e. I can pass in `foo` as a kwarg because there's no validation done.
2. It takes away from immediately readability (you actually /have/ to go through the docs or implementation to figure them out). That also applies to generated docs (also on my TODO list).

> But mixing both doesn't make a ton of sense.

Fair point. I wanted to minimize the surface area of the change, but I think you're right and moving to named kwargs for consistency makes the most sense.

> Also you'll probably want to add something to the docstring about this.

Woops, thanks for the catch.

setup.py
demianbrecht: Ah cool, updated. I've also changed `extra_*_requires` to `pyver_*_requires`. I don't think docs are actually necessary as the name change makes it pretty obvious.

demianbrecht:
Comments should be addressed, will squash before merging.

nickcatal:
Awesome.
Looks good to me.
Thanks

PR #74, opened 2015-09-21
Added support for defining headers in SFType

walterhan:
Took a stab at a quick fix that would allow Sforce-auto-assign to be defined in the headers - I didn't necessarily want to make an argument just for sforce-auto-assign, so I took an approach where you could manually define/change the headers in Salesforce or SFType.

The code change simply allows SFType.headers to be defined and changed. The Salesforce get_attr function also passes in it's own headers to SFType.

To add, for example, Sforce-auto-assign to the headers, you could define it in the Salesforce API class headers and have it pass through to SFType like so:

```
>>> sf = simple_salesforce.Salesforce(sandbox=True, username=username, password=password, security_token=token)
>>> sf.headers
{'X-PrettyPrint': '1', 'Content-Type': 'application/json', 'Authorization': u'Bearer xxxx'}
>>>
>>> sf.headers['Sforce-auto-assign']=False
>>> sf.headers
{'Sforce-auto-assign': False, 'X-PrettyPrint': '1', 'Content-Type': 'application/json', 'Authorization': u'Bearer xxxx'}
>>>
>>> sf.Contact.headers
{'Sforce-auto-assign': False, 'X-PrettyPrint': '1', 'Content-Type': 'application/json', 'Authorization': u'Bearer xxxx'}
>>>

```

Unfortunately, this  makes the SFType headers a function of your API connection headers. If, for example, you wanted to auto assign Cases, but NOT auto assign Leads, you would either have to change your headers before each Case/Lead call, or create an SFType instance for each that defines the headers. The ultimate best solution might be to provide an argument to SFType calls (create, upsert, etc) where you can pass in custom headers.

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -164,7 +164,7 @@ class Salesforce(object):
         if name.startswith('__'):
             return super(Salesforce, self).__getattr__(name)
 
-        return SFType(name, self.session_id, self.sf_instance, self.sf_version, self.proxies)
+        return SFType(name, self.session_id, self.sf_instance, self.sf_version, self.proxies, self.headers)
 
     # User utlity methods
     def set_password(self, user, password):
@@ -394,7 +394,7 @@ class Salesforce(object):
 class SFType(object):
     """An interface to a specific type of SObject"""
 
-    def __init__(self, object_name, session_id, sf_instance, sf_version='27.0', proxies=None):
+    def __init__(self, object_name, session_id, sf_instance, sf_version='27.0', proxies=None, headers=None):
         """Initialize the instance with the given parameters.
 
         Arguments:
@@ -410,6 +410,12 @@ class SFType(object):
         self.name = object_name
         self.request = requests.Session()
         self.request.proxies = proxies
+
+        self.headers = headers if headers else {
+              'Content-Type': 'application/json',
+              'Authorization': 'Bearer ' + self.session_id,
+              'X-PrettyPrint': '1'
+          }
 
         self.base_url = (u'https://{instance}/services/data/v{sf_version}/sobjects/{object_name}/'
                          .format(instance=sf_instance,
@@ -565,12 +571,8 @@ class SFType(object):
 
         Returns a `requests.result` object.
         """
-        headers = {
-            'Content-Type': 'application/json',
-            'Authorization': 'Bearer ' + self.session_id,
-            'X-PrettyPrint': '1'
-        }
-        result = self.request.request(method, url, headers=headers, **kwargs)
+
+        result = self.request.request(method, url, headers=self.headers, **kwargs)
 
         if result.status_code >= 300:
             _exception_handler(result, self.name)
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -405,17 +405,19 @@ class SFType(object):
         * sf_instance -- the domain of the instance of Salesforce to use
         * sf_version -- the version of the Salesforce API to use
         * proxies -- the optional map of scheme to proxy server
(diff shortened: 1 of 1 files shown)


sac2171:
I'd love to get this merged in as well.

simple_salesforce/api.py
```diff
@@ -565,12 +571,8 @@ def _call_salesforce(self, method, url, **kwargs):
...
         Returns a `requests.result` object.
         """
-        headers = {
-            'Content-Type': 'application/json',
-            'Authorization': 'Bearer ' + self.session_id,
-            'X-PrettyPrint': '1'
-        }
-        result = self.request.request(method, url, headers=headers, **kwargs)
+
+        result = self.request.request(method, url, headers=self.headers, **kwargs)
```
sac2171: if self.headers: # use them, 
otherwise: use old.

In case someone was using references to Tasks or Leads for example.

walterhan: Do lines 414-418 take care of that on the __init__ of the sftype? 

        self.headers = headers if headers else {
              'Content-Type': 'application/json',
              'Authorization': 'Bearer ' + self.session_id,
              'X-PrettyPrint': '1'
          }

I can add in a header re-definition/check back into __call_salesforce, just wanted to double check I wasn't missing something.

# contact w/ default headers specified in connection
my_contact = sf.Contact

# leads w/ custom header
sfleads = SFType('Lead', sf.session_id, sf.sf_instance, sf.proxies, mycustomheaders)

simple_salesforce/api.py
```diff
@@ -411,6 +411,12 @@ def __init__(self, object_name, session_id, sf_instance, sf_version='27.0', prox
         self.request = requests.Session()
         self.request.proxies = proxies
 
+        self.headers = headers if headers else {
```
Trii: If you create an `SFType` on your own and pass custom headers you will also need to pass in `Content-Type` and `Authorization` since this change introduces an or instead of a header merge.

nickcatal:
@demianbrecht you got this?

demianbrecht:
@nickcatal: Unfortunately, still waiting on decisions from our legal team.

Apologies for the delay on a response here folks, will get to it as soon as possible.

Thanks for the contribution and efforts!

simple_salesforce/api.py
walterhan: I thought about this, but I opted to have the user define it rather than instituting a merge policy when the same key is redefined/changed in the new header passed in. Probably a little too prudent, and I'd be fine with overwriting the new value for a given key if it's defined in the headers passed in. Thoughts?

Trii: The issue is that if you pass custom headers then you must also remember to
reconstruct the Authorization header as well as the content-type header for
every usage.

walterhan: Just updated code to perform a dict update on original header definition w/ passed in headers. Let me know your thoughts - thanks!

demianbrecht:
Thanks again for the work on this, and my apologies (again) on the delay as we're still working on sorting out legal issues (nothing to worry about for anyone contributing, just investigating options for a potential new home for this project).

> Unfortunately, this makes the SFType headers a function of your API connection headers.

For me, this is a no-go for the patch as-is. According to the docs (https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/headers_autoassign.htm), this is only for Leads and Cases, and only for creates and updates.

Given the current design with the flexible SFType, I would be inclined to agree that the optimal approach would be to allow the addition of custom headers to all calls that result in an HTTP request. I would much prefer that approach to the one here.

demianbrecht:
Closing this due to lack of activity.

PR #75, opened 2015-10-08
Adding ability to modify headers to SimpleSalesforce

sendbloom:
Adding ability to modify headers to SimpleSalesforce


Trii:
User provided headers should be merged into the default headers not replace them or else you will need to pass the auth and content type header as well.

Trii:
@sendbloom you could add a second commit to merge headers instead of closing this PR.

PR #79, opened 2015-10-22
Changed DEFAULT_API_VERSION from '29.0' to '34.0'

babakh:


gcobr:
I'd love to see this merged. I wasted a considerable amount of time to find out why some fields of some sobjects were not being returned in 'describe' calls.

babakh:
Actually, everyone should be at API version '35.0' now with Winter '16 release.  So, it should really be changed to '35.0'.  I have made the change locally and there is no issue.

demianbrecht:
While I'm +1 to the change, I'd like to ensure that there are no backwards incompatible changes in the existing API that current users are dependent on. System tests would be swell ;) I'll dig into this a little more when as soon as I can, unless someone else is able to provide details that give that guarantee.

demianbrecht:
Luckily (?) I just stumbled across a backwards incompatible change released as part of 34.0 that (IMO), makes this change a no-go until a major version release: http://releasenotes.docs.salesforce.com/en-us/summer15/release-notes/rn_api_error_codes.htm.

Because the older error has been replaced by newer (more readable) ones, it would break existing code.

I'm going to close this for now, given I'm expecting that another release may happen before we get to a major release of simple_salesforce, but will add a TODO item to the issue tracker with a new 1.0 milestone to update the API version to the most recent at release time.

PR #82, opened 2015-11-09
Fix version param

demianbrecht:
The version param of Salesforce.__init__ was changed to sf_version. This
patch changes it back.


demianbrecht:
@nickcatal: Mind taking a quick look? This fixes a backwards incompatible change I introduced.

FYI @shawnatsalesforce.

nickcatal:
This looks right to me.

PR #85, opened 2015-11-14
Add tox support

demianbrecht:
Add tox support to enable local test runs matching those on Travis-CI

* Added tox support
* Broke up unit tests and static analysis
* Build will now break if static analysis fails
* Build will now break if code coverage drops below 62%
* Changes made to ensure pylint passes


demianbrecht:
@nickcatal: Thoughts on this change?

nickcatal:
Love it. Looks good to me.

PR #89, opened 2015-11-26
Fix defining of status and resource_name for SalesforceGeneralError

maxvyaznikov:
Subj.
In other case, in raised SalesforceGeneralError exception, we have status with value of resource_name, and resource_name with value of status.


demianbrecht:
Thanks for the patch @maxbeizer and apologies for the _long_ delay in a response. Rather than use named params everywhere, can't the order of parameters in the problematic `SalesforceGeneralError`s be changed to match what's expected by `simple_salesforce.util.SalesforceError`?

andscoop:
This should be handled by #207. Can close when/if that is merged.

PR #97, opened 2016-02-15
Use _call_salesforce everywhere

### LINKED ISSUE #96: Expose methods other than GET via restful API
Currently, `Salesforce.restful` only allows for `get` API calls. There should be a mechanism that allows a user to use other HTTP methods as well.

Reported by @babakh.

Trii:
This fixes #96 without breaking BC by allowing an http method to be passed as a named 3rd parameter to `restful`.

Also uses `_call_salesforce` everywhere for consistency.


demianbrecht:
Thanks for the contribution!

PR #98, opened 2016-02-15
Share requests.Session across SFType instances

### LINKED ISSUE #91: SFType doesn't inherit session.
This effectively disables connection pooling and greatly hurts performance.

Trii:
Fixes #91 

I was a bit more explicit in the creation of `SFType` and specified parameter names instead of relying on their position. I hope that is OK.

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -73,7 +73,10 @@ class Salesforce(object):
         # kwargs
         self.sf_version = version
         self.sandbox = sandbox
-        self.proxies = proxies
+        self.request = session or requests.Session()
+        self.proxies = self.request.proxies
+        if proxies is not None:  # override custom session proxies dance
+            self.request.proxies = self.proxies = proxies
 
         # Determine if the user wants to use our username/password auth or pass
         # in their own information
@@ -83,7 +86,7 @@ class Salesforce(object):
 
             # Pass along the username/password to our login helper
             self.session_id, self.sf_instance = SalesforceLogin(
-                session=session,
+                session=self.request,
                 username=username,
                 password=password,
                 security_token=security_token,
@@ -110,7 +113,7 @@ class Salesforce(object):
 
             # Pass along the username/password to our login helper
             self.session_id, self.sf_instance = SalesforceLogin(
-                session=session,
+                session=self.request,
                 username=username,
                 password=password,
                 organizationId=organizationId,
@@ -129,8 +132,6 @@ class Salesforce(object):
         else:
             self.auth_site = 'https://login.salesforce.com'
 
-        self.request = session or requests.Session()
-        self.request.proxies = self.proxies
         self.headers = {
             'Content-Type': 'application/json',
             'Authorization': 'Bearer ' + self.session_id,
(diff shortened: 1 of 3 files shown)


demianbrecht:
Hey @Trii, cheers for the patch!

Can you perhaps add a small unit test or two confirming the expected functionality? Offhand what I'm looking for specifically is to ensure that an instance of `SFType` is reusing the `requests.Session` of the `Salesforce` instance that's instantiating it.

Otherwise, LGTM. Unfortunately, I'm still blocked on merging PRs, but hoping to have that cleared up ASAP.

Thanks again!

Trii:
Will do!

Trii:
@demianbrecht tests for sharing a session and making sure proxies propagate through the `__getattr__`  calls are set. I also noticed that when `session=None`, `SalesforceLogin` was using `requests.post` but then afterwards, `Salesforce` was creating a new `requests.Session` and using it from then on out. I moved the logic up so that a new `requests.Session` is being passed to `SalesforceLogin` for consistency. [This change here](https://github.com/heroku/simple-salesforce/pull/98/files#diff-386c8e7f4cc62563b2b586356d1c67ffR76)

simple_salesforce/api.py
```diff
@@ -442,11 +443,15 @@ def __init__(
...
         * proxies -- the optional map of scheme to proxy server
+        * session -- Custom requests session, created in calling code. This
+                     enables the use of requests Session features not otherwise
+                     exposed by simple_salesforce.
         """
         self.session_id = session_id
         self.name = object_name
-        self.request = requests.Session()
-        self.request.proxies = proxies
+        self.request = session or requests.Session()
```
demianbrecht: I see the inconsistent naming between this and the `Salesforce` object (`request` here but `session` in `Salesforce`). I would much rather see `session` used consistently between the two as it's what it's actually a reference to an instance of. Also, parameters and attributes will also be consistent.

simple_salesforce/api.py
```diff
@@ -442,11 +443,15 @@ def __init__(
...
+        if proxies is not None:  # don't wipe out original proxies with None
+            self.request.proxies = proxies
```
demianbrecht: I think that if a session is passed in explicitly, `proxies` should be ignored. The side effect on the session instance owned by an upstream object concerns me a little.

demianbrecht:
Hey @Trii, we're now unblocked on accepting contributions! I've made a couple comments on the PR, hopefully you're still interested in contributing after such a long time.

simple_salesforce/api.py
demianbrecht: Note to self: Stop being off your face.

Trii:
Hey @demianbrecht, I don't work with Salesforce any longer but if you'd like to merge it still I can address the comments and fix up the PR

demianbrecht:
I was going to pick this up myself, but if you have the cycles to finish 'er up, that'd be awesome. Cheers!

Trii:
Should be all set @demianbrecht. I ended up rebasing off of the current master since this PR was so old.

demianbrecht:
Thanks for the quick turnaround @Trii! Made a couple slight modifications but great stuff.

Trii:
You're welcome. Hope it helps!

PR #104, opened 2016-03-24
Allow custom HTTP headers when calling the update() method

nitzanm:
There are some cases where the user needs to pass a custom HTTP header when making a record update request.  The situation I ran into was passing the custom "Sforce-Auto-Assign" header when updating a lead to prevent auto-assign rules from running:

https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/headers_autoassign.htm

This trivial library change allows the user to pass custom headers; for example, in the situation above:

    sf.Lead.update(lead_id, lead_updates, headers={'Sforce-Auto-Assign': 'FALSE'})

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -543,7 +543,7 @@ class SFType(object):
                                        data=json.dumps(data))
         return self._raw_response(result, raw_response)
 
-    def update(self, record_id, data, raw_response=False):
+    def update(self, record_id, data, raw_response=False, **kwargs):
         """Updates an SObject using a PATCH to
         `.../{object_name}/{record_id}`.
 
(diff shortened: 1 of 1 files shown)


meilke:
:+1:

demianbrecht:
Thanks for the contribution @nitzanm and my apologies for taking so long to get to this!

Is there a reason to use `**kwargs` rather than just adding an explicit `headers` named param? The latter is much more obvious.

Also, there should be tests accompanying the change.

tarxvf:
👍👍

GlenRSmith:
LGTM

bfernandezsp:
👍

meilke:
I created a pull request at https://github.com/simple-salesforce/simple-salesforce/pull/131 that is the same changes on a newer fork including basic tests.

demianbrecht:
Closing this due to the additional work in #131.

PR #106, opened 2016-03-27
add support for execute anonymous via tooling api

realdoug:
I had a need to execute anonymous apex, so I added support for the tooling endpoint and the executeAnonymous resource.  Thanks!


jlecount-sungevity:
Any down-side to merging this?  Looks like functionality we'd also use...

simple_salesforce/api.py
```diff
@@ -140,6 +140,24 @@ def __init__(
                                  version=self.sf_version))
         self.apex_url = ('https://{instance}/services/apexrest/'
                          .format(instance=self.sf_instance))
+        self.tooling_url = ('{base}tooling/').format(base=self.base_url)
+
+    def execanon(self, apex_string):
```
harmon: Maybe `exec_anon`, to match the naming of other methods. Looks good to me!

simple_salesforce/api.py
```diff
@@ -140,6 +140,24 @@ def __init__(
...
+        """Executes a string of Apex code.
+        """
+        url = self.tooling_url + "executeAnonymous/"
+        params = {'anonymousBody': apex_string}
+        result = self.request.get(url, headers=self.headers, params=params)
```
demianbrecht: This should use `_call_salesforce` for consistency.

demianbrecht:
Apologies for the delay getting back to this. Tests should be added and `_call_salesforce` should be used rather than `requests` directly. Also, I think `apexecute_anon` would likely be a better name given the existence of `apexecute`.

simple_salesforce/api.py
Daenyth: 👍  to that; I have to override `_call_salesforce` in my codebase so that I can hook into some request-tracking code in my codebase. The consistency made it really easy

Issue #103, opened 2016-03-08
Salesforce and TLSv1.1


calpaterson:
Soon [Salesforce will disable TLSv1](https://help.salesforce.com/apex/HTViewSolution?id=000221207).  

A surprising number of people are still using versions of Python before 2.7.9, which is the first version of Python 2 to support TLSv1.1.  This includes users of the most recent Ubuntu LTS, RHEL 7 and many other popular linux distros.  These users will need to have PyOpenSSL and friends installed (with a system openssl library >= 1.0.1) so that the requests library simple_salesforce uses can speak TLSv1.1.

Could you include documentation for these users to at least give them a pointer about where to go if they have this problem?  Are PRs on this topic welcome?

cyndiec:
@calpaterson I've run into this issue, and not sure where to go from here. Any ideas?

`in SalesforceLogin
simple_salesforce.login.SalesforceAuthenticationFailed: UNSUPPORTED_CLIENT: TLS 1.0 has been disabled in this organization. Please use TLS 1.1 or higher when connecting to Salesforce using https.`

calpaterson:
@cyndiec My understanding is that Salesforce as a whole hasn't disabled TLS 1.0 yet so your administrator might have changed a setting.  Perhaps you can persuade them to change it back!  Alternatively you can install the `requests[security]` package which will install and compile libraries that allow your system openssl to be used.  Hopefully your system openssl new enough to allow for TLSv1.1 (needs to be 1.0.1 or later)

@demianbrecht Ok.  When/if you can accept contributions again I'm happy to write this up.  Maybe my organisation has some other stuff we can contribute back too - I'll ask.

cyndiec:
@calpaterson Our administrator ended up switching it back for now. Thank you so much for taking the time to explain this to me. I appreciate it.

demianbrecht:
Ah yes, I'd forgotten about the `[security]` option for `requests` (I ended up installing `ndg-httpsclient` myself after debugging and finding that urllib3 uses it automatically if available). Thanks for pointing that out @calpaterson!

demianbrecht:
@calpaterson we're good to accept contributions again, so if you have the bandwidth, a PR is most welcome!

demianbrecht:
Should be resolved by #110 and available in v0.69.

PR #110, opened 2016-06-09
Updated requests dependency to requests[security]

demianbrecht:
Addresses issue #103

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,13 +1,14 @@
-v0.69 (Release date: TBD)
-=========================
+v0.69
+=====
 
 Features
 --------
 - [103] Require requests[secure] to allow for system openssl, which will allow
   for TLSv1.2 in Python < 2.7.9 assuming system openssl is newer than Python's.
 
-Bug Fixes
----------
+
+v0.68.2
+=======
 
 Other
 -----
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,11 @@ Features
 - [103] Require requests[secure] to allow for system openssl, which will allow
   for TLSv1.2 in Python < 2.7.9 assuming system openssl is newer than Python's.
 
+Other
+-----
+- Replaced httpretty with responses for unit tests due to SSL-related errors
+  in httpretty.
+
 
 v0.68.2
 =======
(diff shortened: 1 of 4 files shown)


demianbrecht:
FYI @nickcatal

nickcatal:
Can you update the version in setup.py so we can send to pypi?

nickcatal:
Q2: Would it make sense to use `pyver_install_requires` so that people on newer python aren't required to download a new version of requests from what they may already have installed either in their existing simple-salesforce environment or system-wide?

I'm inclined to think that makes more sense. Especially since we do test against older versions of python.

Thoughts?

demianbrecht:
> Can you update the version in setup.py so we can send to pypi?

Yep, will do.

> Would it make sense to use pyver_install_requires so that people on newer python aren't required to download a new version of requests from what they may already have installed either in their existing simple-salesforce environment or system-wide?

I was considering that. My thinking here was that, because openssl releases more frequently than python, it would likely be best from a security perspective to allow for system openssl use across the board instead of only < 2.7.9.

That said, it can also have adverse effects should the user's system openssl be older than Python's (hopefully never the case, but I'm sure it's bound to happen). Perhaps a disclaimer in the README about ensuring an up to date system openssl would be the best way to mitigate any backwards incompatibility issues that crop up due to the change?

nickcatal:
@demianbrecht it looks like [2.6](https://travis-ci.org/simple-salesforce/simple-salesforce/jobs/136519654), [2.7](https://travis-ci.org/simple-salesforce/simple-salesforce/jobs/136519656), and pypi are failing tests with this due to handshake issues.

I wasn't aware we were doing actual calls to salesforce as part of tests, but any idea what might be going on here? I can't think of anything off the top of my head.

demianbrecht:
It's an issue with httpretty (it's in the callstack). The issue in their issue tracker: https://github.com/gabrielfalcao/HTTPretty/issues/242. I've run into problems with httpretty before, so it might be worthwhile to just swap it out for responses (https://github.com/getsentry/responses). I'm gonna give that a go.

demianbrecht:
All tests are passing, @nickcatal when you get a chance, I'll merge this after a 👍

nickcatal:
Looks good to me!

PR #119, opened 2016-07-01
Expose the api limits and usage header from the response

### LINKED ISSUE #118: Expose API usage
The Salesforce API returns a usage report in the header of each response, as detailed in the documentation: https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/headers_api_usage.htm
it looks something like `Sforce-Limit-Info: api-usage=18/5000`. 

It would be very helpful to be able to track our API usage, but it doesn't look like the simple-salesforce library exposes this information. I'd like to add a variable to instances of `Salesforce` objects that gets updated in `_call_salesforce()`, and potentially a similar variable in `SFType` objects that transfers to the `Salesforce` object in the `__getattr__` method. Are there any other methods I should account for in this change?

cseager:
This resolves #118 by recording api usage so that we can monitor it without needing to make extra requests.

Example usage: 

``` python
>>> import simple_salesforce
>>> salesforce_client = simple_salesforce.Salesforce( <credentials> )
>>> a = salesforce_client.Account
>>> q = 'SELECT SystemModstamp FROM Account WHERE SystemModstamp > 2016-06-20T00:00:00Z LIMIT 10'
>>> _ = salesforce_client.query(q)
>>> salesforce_client.api_usage
{'api-usage': Usage(used=18, total=5000)}
>>> _ = a.describe()
>>> a.api_usage
{'api-usage': Usage(used=19, total=5000)}
```

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -7,6 +7,7 @@ DEFAULT_API_VERSION = '29.0'
 
 import requests
 import json
+import re
 
 try:
     from urlparse import urlparse
@@ -140,6 +141,8 @@ class Salesforce(object):
                                  version=self.sf_version))
         self.apex_url = ('https://{instance}/services/apexrest/'
                          .format(instance=self.sf_instance))
+
+        self.api_usage = {}
 
     def describe(self):
         """Describes all available objects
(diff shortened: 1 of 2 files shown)


simple_salesforce/api.py
```diff
@@ -421,8 +424,40 @@ def _call_salesforce(self, method, url, **kwargs):
...
+
+        Arguments:
+
+        * sforce_limit_info: The value of response header 'Sforce-Limit-Info'
+            Example 1: 'api-usage=18/5000'
+            Example 2: 'api-usage=25/5000;
+                per-app-api-usage=17/250(appName=sample-connected-app)'
+        """
+        result = {
+            'Sforce-Limit-Info': sforce_limit_info
```
demianbrecht: If the values are being deserialized, why return the raw value as well?

simple_salesforce/api.py
```diff
@@ -421,8 +424,40 @@ def _call_salesforce(self, method, url, **kwargs):
...
+        }
+
+        api_usage = re.match(r'[^-]?api-usage=(?P<used>\d+)/(?P<tot>\d+)',
+                             sforce_limit_info)
+        pau = r'.+per-app-api-usage=(?P<u>\d+)/(?P<t>\d+)\(appName=(?P<n>.+)\)'
+        per_app_api_usage = re.match(pau, sforce_limit_info)
+
+        if api_usage and api_usage.groups():
+            result['api-usage'] = [int(n) for n in  api_usage.groups()]
```
demianbrecht: Rather than a bare list of numbers, I think I'd rather see something with a little more context. Maybe `namedtuple`s for both global and per-app usage?

demianbrecht:
Thanks for the contribution @cseager, apologies for the delayed response. Just a couple comments left in the PR.

cseager:
@demianbrecht I've made some updates based on your suggestions. Thanks for taking the time to look!

kenoung:
what happened to this pull request? are there plans for this functionality to be added?

PR #121, opened 2016-07-07
Readme: Use keyword arguments in login example

johbo:
Was trapped by the positional arguments example, noticed that login
supports organization id as well as security token. Using keyword
arguments should avoid this trap.


demianbrecht:
Thanks for the submission!

PR #124, opened 2016-08-03
Add Client Id To Identify API Source

rynmlng:
Adding a client Id to the SOAP Header will allow API calls in Salesforce to be grouped by client Id, and thus identifiable when viewing the "breakdown of API calls." This is great when you need to analyze how many API calls each of your apps are making, especially projecting if you need to adjust your maximum API requests associated with your account.

Currently when you view a report on recent API calls segmented by Client Id, calls made by `simple_salesforce` are identified by a Client Id of `-`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -58,6 +58,15 @@ For example:
 
 Note that specifying if you want to use a sandbox is only necessary if you are using the built-in username/password/security token authentication and is used exclusively during the authentication step.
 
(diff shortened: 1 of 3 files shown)


rynmlng:
All checks pass! <3 <3 <3

nickcatal:
This is smart. Thanks!

Issue #112, opened 2016-06-14
Split services


demianbrecht:
There's currently a mix of services implemented directly in `simple_salesforce.Salesforce` (i.e. apex and sobjects). This should be changed such that `simple_salesforce.Salesforce` contains references to service implementations, such as `Salesforce.bulk`, `Salesforce.sobjects`, `Salesforce.reports`, `Salesforce.apex`, etc.

As a part of this, the existing APIs should be deprecated (use `warnings.warn(..., DeprecationWarning)`) and removed in a future version.

nickcatal:
When you say `sobject` are you saying the record management functionality currently done via attributes?

What do you see as the future way of changing and updating records? `sf.Contacts.update()` seems pretty straightforward.

Thanks

demianbrecht:
> What do you see as the future way of changing and updating records? sf.Contacts.update() seems pretty straightforward.

It's straightforward when you only consider the force.com API, but starts to get a little unclear when you consider adding support for other services (https://help.salesforce.com/HTViewHelpDoc?id=integrate_what_is_api.htm).

What I'm proposing is simply something like this:

```
sf.force.Contacts.update()
sf.chatter.do_something()
sf.tooling.do_something()
sf.apex.do_something()
sf.bulk.do_something()
```

It's much more representative of the backend services that they're implementing and allows for modularity between services. It would also allow for special per-service handling independently of other service implementations. In a nutshell, it's the separation of concerns I'm after with this change. `Salesforce` should manage connectivity and any common aspects (error handling, serialization and deserialization, etc), but the actual service implementation can be (and *should* be) independent of one another.

nickcatal:
This looks great! And I think writing the relevant depreciation warnings would be rather easy as well, which is nice.

ddow:
Is there any update on this?  I'm about a medium level in Python but I use it every day to access salesforce in any number of ways so if you need someone to be a guinea pig any run your code I'd be more than willing and you get a lot of feedback.  I really need this and I can't find a way to do it with Python 3.  I use beatbox in Python 2, but it does not seem that it is being developed anymore.

colemanja91:
I also would be willing to contribute development/testing time.

demianbrecht:
@colemanja91, @ddow would be more than happy to get and review PRs related to this change. I've been on a bit of a hiatus, but will be more active going forward.

demianbrecht:
[...]

PR #127, opened 2016-08-08
Bulk API Support

prathik:
Please review Bulk API support added for Salesforce

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -445,6 +445,10 @@ class Salesforce(object):
         return result
 
     def _call_bulk_api(self, method, url, **kwargs):
+        """Makes a call to the bulk api
+
+        Returns a `request.result` object.
+        """
         result = self.request.request(
             method, url, headers=self.bulk_headers, **kwargs)
 
@@ -454,14 +458,23 @@ class Salesforce(object):
         return result
 
     def _make_bulk_request(self, url, data):
+        """Wrapper function to call a bulk API.
+
+        Parses the response to a dict.
+
+        Returns a `request.result` and a parsed
+        `reponse_object` dict
+        """
         response = self._call_bulk_api("POST", url, data=data)
         response_object = xmltodict.parse(response.text)
         return response, response_object
 
     def _template_location(self, template_file_name):
+        """Location of the template file"""
         return os.path.join(self.location, template_file_name)
 
     def _close_job(self, job_id):
+        """Makes Salesforce call to close a job"""
         with open(self._template_location('close_job.xml'),
                   'r') as close_xml_file:
             close_xml_string = close_xml_file.read()
@@ -479,22 +492,25 @@ class Salesforce(object):
                                              response.text)
 
     def _create_job(self, data):
+        """Makes a bulk api request to create a job"""
         url = self.bulk_url
         job_info = self._make_bulk_request(url, data)
+
         return job_info[1]['jobInfo']['id']
 
-    def _create_basic_job(self, object_name, type):
-        with open(self._template_location(
-                'basic_job_xml_template.xml'), 'r') as \
-                basic_job_template_file:
+    def _create_basic_job(self, object_name, method_type):
+        """Create a basic job, eg. insert"""
+        with open(self._template_location('basic_job_xml_template.xml'),
+                  'r') as basic_job_template_file:
             basic_job_template_string = basic_job_template_file.read()
-            basic_job_xml = basic_job_template_string.format(type, object_name)
+            basic_job_xml = basic_job_template_string.format(method_type,
+                                                             object_name)
             return self._create_job(basic_job_xml)
 
     def _create_upsert_job(self, object_name, external_field):
-        with open(self._template_location(
-                'upsert_job_xml_template.xml'), 'r') as \
-                upsert_job_xml_template_file:
+        """Create an upsert job"""
+        with open(self._template_location('upsert_job_xml_template.xml'),
+                  'r') as upsert_job_xml_template_file:
             upsert_job_template_string = upsert_job_xml_template_file.read()
             upsert_job_xml = upsert_job_template_string.format(
                 object_name,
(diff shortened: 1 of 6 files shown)


prathik:
@demianbrecht Please review also would be great if you could suggest how we should go about adding tests for this

simple_salesforce/api.py
```diff
@@ -75,6 +78,10 @@ def __init__(
         self.sandbox = sandbox
         self.proxies = proxies
 
+        self.location = os.path.realpath(
+            os.path.join(os.getcwd(), os.path.dirname(__file__)))
```
demianbrecht: This will only work for test cases where the code's being executed in the project dir, not for dependent projects. That said, I'd personally prefer to see the templates inline in this file given it's really minimal.

simple_salesforce/api.py
```diff
@@ -135,9 +142,23 @@ def __init__(
             'X-PrettyPrint': '1'
         }
 
+        self.bulk_headers = {
```
demianbrecht: Is there a reason to have these as public attributes? I know it's consistent with `self.headers`, but do you foresee them being used external to this library? If not, I think I'd prefer to see these as literals in `_call_bulk_api`.

simple_salesforce/api.py
```diff
@@ -135,9 +142,23 @@ def __init__(
...
+            'Content-Type': 'application/xml',
+            'X-SFDC-Session': self.session_id,
+            'charset': 'UTF-8'
+        }
+
+        self.batch_headers = {
```
demianbrecht: Same thing here about public availability rather than internal to `_create_batch`

simple_salesforce/api.py
```diff
@@ -423,6 +444,115 @@ def _call_salesforce(self, method, url, **kwargs):
...
+        """
+        result = self.request.request(
+            method, url, headers=self.bulk_headers, **kwargs)
+
+        if result.status_code >= 300:
+            _exception_handler(result)
+
+        return result
+
+    def _make_bulk_request(self, url, data):
```
demianbrecht: If this is the only entry point to `_call_bulk_api`, is there a reason to have this extra function given it's only 3 lines?

simple_salesforce/api.py
```diff
@@ -423,6 +444,115 @@ def _call_salesforce(self, method, url, **kwargs):
...
+        Parses the response to a dict.
+
+        Returns a `request.result` and a parsed
+        `reponse_object` dict
+        """
+        response = self._call_bulk_api("POST", url, data=data)
+        response_object = xmltodict.parse(response.text)
+        return response, response_object
+
+    def _template_location(self, template_file_name):
```
demianbrecht: Unneeded if the template content is inline (no reason for I/O here)

simple_salesforce/api.py
```diff
@@ -423,6 +444,115 @@ def _call_salesforce(self, method, url, **kwargs):
...
+        """Location of the template file"""
+        return os.path.join(self.location, template_file_name)
+
+    def _close_job(self, job_id):
+        """Makes Salesforce call to close a job"""
+        with open(self._template_location('close_job.xml'),
+                  'r') as close_xml_file:
+            close_xml_string = close_xml_file.read()
+            url = self.bulk_url + "/" + job_id
```
demianbrecht: Nitpick: I prefer interpolation over addition for strings (i.e. `'foo %s' % bar`)

simple_salesforce/api.py
```diff
@@ -423,6 +444,115 @@ def _call_salesforce(self, method, url, **kwargs):
...
+                                             "close",
+                                             response.status_code,
+                                             response.text)
+
+    def _create_job(self, data):
+        """Makes a bulk api request to create a job"""
+        url = self.bulk_url
+        job_info = self._make_bulk_request(url, data)
+
+        return job_info[1]['jobInfo']['id']
```
demianbrecht: Why the first element? At least a comment explaining this would be helpful.

simple_salesforce/api.py
```diff
@@ -423,6 +444,115 @@ def _call_salesforce(self, method, url, **kwargs):
...
+            upsert_job_xml = upsert_job_template_string.format(
+                object_name,
+                external_field)
+            return self._create_job(upsert_job_xml)
+
+    def _create_batch(self, csv_file, job_id):
+        """Create a batch from a csv file"""
+        csv_file_data = open(csv_file, 'rb').read()
+        url = self.bulk_url + "/" + job_id + "/batch"
+        self.request.request(
```
demianbrecht: What if this fails for any reason? Should there not be any error handling here?

simple_salesforce/api.py
```diff
@@ -423,6 +444,115 @@ def _call_salesforce(self, method, url, **kwargs):
...
+        with open(self._template_location('close_job.xml'),
+                  'r') as close_xml_file:
+            close_xml_string = close_xml_file.read()
+            url = self.bulk_url + "/" + job_id
+            response, \
+                response_object = self._make_bulk_request(url,
+                                                          close_xml_string)
+
+            if 'jobInfo' not in response_object or \
+               'state' not in response_object['jobInfo'] or \
```
demianbrecht: Indentation seems off here. This should be two additional levels deeps, cleanly separating it from the block's content.

simple_salesforce/api.py
```diff
@@ -423,6 +444,115 @@ def _call_salesforce(self, method, url, **kwargs):
...
+        """Bulk upload a CSV to a Salesforce Object
+
+        Arguments:
+
+        * object_name -- name of the object to upload to
+        * csv_file -- data to upload to the object"""
+
+        job_id = self._create_basic_job(object_name, "insert")
+        self._create_batch(csv_file, job_id)
+        self._close_job(job_id)
```
demianbrecht: Should jobs *always* be closed? What if there's a socket error or something unexpected that happened in `_create_batch`? Should this perhaps be in a try/except block?

demianbrecht:
Thanks for the work! I've done a code-level review and have left a few comments (I'll take a look at the actual bulk/batch APIs as soon as I can as well). Tests will also need to be added.

I'd *love* to see this broken up into sub services as outlined in #112 before we add more APIs that will require backwards compatible support (at least for minor revisions). But obviously that's quite a bit more work than what you're after getting done here so I'm (personally) okay with merging once comments are addressed. @nickcatal any thoughts here?

nickcatal:
I'm fine with merging it in after comments are addressed. No sense in
delaying all this work if we don't have a timetable for the refactoring.

PR #131, opened 2016-09-14
Allow custom HTTP headers when calling the update() method

meilke:
**Same changes as in https://github.com/simple-salesforce/simple-salesforce/pull/104 but with tests.**

## Original Description

There are some cases where the user needs to pass a custom HTTP header when making a record update request. The situation I ran into was passing the custom "Sforce-Auto-Assign" header when updating a lead to prevent auto-assign rules from running:

https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/headers_autoassign.htm

This trivial library change allows the user to pass custom headers; for example, in the situation above:

    sf.Lead.update(lead_id, lead_updates, headers={'Sforce-Auto-Assign': 'FALSE'})

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ dist/
 .tox/
 tags
 .env/
+.tags
+.tags_sorted_by_file
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -11,10 +11,10 @@ import requests
 import json
 
 try:
-    from urlparse import urlparse
+    from urlparse import urlparse, urljoin
 except ImportError:
     # Python 3+
-    from urllib.parse import urlparse
+    from urllib.parse import urlparse, urljoin
 from simple_salesforce.login import SalesforceLogin
 from simple_salesforce.util import date_to_iso8601, SalesforceError
 
@@ -513,7 +513,8 @@ class SFType(object):
         * headers -- a dict with additional request headers.
         """
         result = self._call_salesforce(
-            'GET', self.base_url + 'describe', headers=headers
+            method='GET', url=urljoin(self.base_url, 'describe'),
+            headers=headers
         )
         return result.json(object_pairs_hook=OrderedDict)
 
@@ -529,9 +530,12 @@ class SFType(object):
         * record_id -- the Id of the SObject to get
         * headers -- a dict with additional request headers.
         """
+        custom_url_part = 'describe/layouts/{record_id}'.format(
+            record_id=record_id
+        )
         result = self._call_salesforce(
-            'GET',
-            self.base_url + 'describe/layouts/' + record_id,
+            method='GET',
+            url=urljoin(self.base_url, custom_url_part),
             headers=headers
         )
         return result.json(object_pairs_hook=OrderedDict)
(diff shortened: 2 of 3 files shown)


meilke:
@demianbrecht, I added a more strict `headers` parameter to all the `SFType` methods. Also added simple headers/no headers tests for all of these.

simple_salesforce/api.py
```diff
@@ -493,43 +493,61 @@ def __init__(
...
         """Returns the result of a GET to `.../{object_name}/describe` as a
         dict decoded from the JSON payload returned by Salesforce.
+
+        Arguments:
+
+        * headers -- a dict with additional request headers.
         """
-        result = self._call_salesforce('GET', self.base_url + 'describe')
+        result = self._call_salesforce(
+            'GET', self.base_url + 'describe', headers=headers)
```
demianbrecht: Nitpicky, but I'd prefer to see either interpolation or `urljoin` used here instead of addition (even though it's used elsewhere): `'%s/%s' % (self.base_url, 'describe',)`

simple_salesforce/api.py
```diff
@@ -662,6 +691,8 @@ def _call_salesforce(self, method, url, **kwargs):
             'Authorization': 'Bearer ' + self.session_id,
             'X-PrettyPrint': '1'
         }
+        additional_headers = kwargs.pop('headers', dict())
+        headers.update(additional_headers or dict())
```
demianbrecht: You don't need `additional_headers` assignment. You can get away with doing it in a single line:

`headers.update(kwargs.pop('headers', dict()))`

The way you have it right now, if `headers` isn't a key in `kwargs`, the default value of `dict()` will be returned. Then in the `headers.update` call, the empty dict will evaluate to be Falsy, which will incur another needless `dict()` creation. Okay, so I'm being a little overly pedantic, but the one-liner is a little clearer to me.

simple_salesforce/tests/test_api.py
```diff
@@ -29,10 +30,433 @@
...
-    SalesforceGeneralError
+    SalesforceGeneralError,
+    SFType
 )
 
 
+def _create_sf_type():
+    """Creates SFType instances"""
+    return SFType(
+        object_name='Case',
```
demianbrecht: Might it be worthwhile to have these attributes configurable through input parameters?

demianbrecht:
Minor comments remaining, not all of which are actually required to get this merged if you're opposed to them enough. Thanks for the work!

.gitignore
```diff
@@ -9,3 +9,5 @@ dist/
 .tox/
 tags
 .env/
+.tags
```
meilke: I would have to change all my other repos... At the end of the day it is two more file names in the .gitignore.

simple_salesforce/tests/test_api.py
meilke: Done!

simple_salesforce/api.py
meilke: Done!

meilke:
@demianbrecht Yet another round of commits ;-)

demianbrecht:
Thanks for the contribution @meilke!

jjfalling:
Do you know when you will be able to release a new version to PyPI that includes this PR @demianbrecht ?

demianbrecht:
@jjfalling I was hoping to get another couple PRs in this release, but I've fallen a little short on time. 0.71 is up now with this change.

jjfalling:
Awesome, thank you!

PR #133, opened 2016-10-01
modify total size correctly

ukisoft:
Result totalSize is more than actual total size when we need to retrieve the remaining SOQL query results.
I think we don't need to add totalSize each retrieving.


demianbrecht:
@nickcatal can you remember why this `totalSize` is increased on every iteration? I couldn't find super clear docs on it, but it /seems/ that it should just be a static number.

nickcatal:
@demianbrecht No idea. Looks like that was introduced in cf86f3bd61cd1f28ac452e93cdd8a65dc96ba270 and just never reviewed

demianbrecht:
Closing this one as it was included in #134.

Issue #132, opened 2016-09-30
Handle large result sets


shreyasjoshis:
### Problem
Large result sets (100K+) cause `RecursionError: maximum recursion depth exceeded while calling a Python object` due to [this recursion](https://github.com/simple-salesforce/simple-salesforce/blob/master/simple_salesforce/api.py#L407).

### Solution
* Use a non recursive implementation to get results
* Use a generator to process results in small batches rather than holding the entire result set in memory

PR #134, opened 2016-10-05
Support large result sets

shreyasjoshis:
Per #132, the recursive implementation was hitting the recursion limit for a large result set. Replace with a generator.

Note that in its current form this will break for people iterating over the result set multiple times.

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -370,8 +370,8 @@ class Salesforce(object):
         return result.json(object_pairs_hook=OrderedDict)
 
     def query_all(self, query, **kwargs):
-        """Returns the full set of results for the `query`. This is a
-        convenience wrapper around `query(...)` and `query_more(...)`.
+        """Returns a generator for the full set of results for the `query`.
+        This is a convenience wrapper around `query(...)` and `query_more(...)`.
 
         The returned dict is the decoded JSON payload from the final call to
         Salesforce, but with the `totalSize` field representing the full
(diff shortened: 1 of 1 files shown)


demianbrecht:
I'm not overly keen on this as is due to the backwards incompatibility as you noted. Additionally, using the generator, per-iteration overhead will be incurred on page exhaustion rather than at call-time. That said, I can't see that really being a problem.

I don't see why the recursion was actually needed. Why not take what you have in the generator and turn it into an iterative implementation instead? That'll get around both the backwards incompatibility problem as well as the recursion limit.

Also, as a /horrible/ alternative, you could always increase the recursion limit using `sys.setrecursionlimit` ;)

shreyasjoshis:
Done

simple_salesforce/api.py
```diff
@@ -383,34 +384,21 @@ def query_all(self, query, **kwargs):
...
+        while True:
+            all_records.extend(result['records'])
+            # fetch next batch if we're not done else break out of loop
+            if not result['done']:
+                result = self.query_more(result['nextRecordsUrl'],
+                                         True)
+            else:
+                break
+
+        result['records'] = all_records
```
shreyasjoshis: This is an attempt at backward compatibility, since this is basically what the code was doing. 

Not sure why the `done` or `totalSize` fields need to be returned, since `done` will always be `True` and `totalSize` should always be equal to `len(result['records'])` here. 

Note that unlike the original implementation and per [this PR](https://github.com/simple-salesforce/simple-salesforce/pull/133) I am simply returning the `totalSize`, which seems to be the right thing to do.

shreyasjoshis:
@demianbrecht is this 👍?

kclulow:
Possible to have this merged to master soon? This is causing my project some pain...

demianbrecht:
@shreyasjoshis apologies for the delay, I've been insanely busy of late. I'll get this merged today, LGTM.

demianbrecht:
Thanks for the work @shreyasjoshis!

PR #140, opened 2016-11-23
add bulk_url to Salesforce instance; sets up for future features arou…

colemanja91:
add bulk_url to Salesforce instance; sets up for future features around Bulk/Async API.


colemanja91:
rolling this into next PR with full bulk API functionality.

PR #143, opened 2016-12-21
Feature/bulk api

colemanja91:
Adding support for Bulk API according to discussions in #112 . 

- Split out bulk classes into `bulk.py` as a proposed model for splitting other services; this way, the handling classes for each API type would live in it's own file, making the code more manageable. 
- Added example docs to README
- I've tested in the sandbox available to me, with a few object types. 

Potential edits/discussion: 

- To get the format mentioned in #112, `sf.bulk.Contact.do_something()`, had to create two new classes, `SFBulkHandler` and `SFBulkType`. I'm not entirely happy with `SFBulkHandler` as a solution, since it's only purpose is to parse out the SF object type and pass to `SFBulkType`. If anyone has ideas on how to better handle this, let me know and I'll work it in. Another alternative is to use the format `sf.bulk.do_something(object='Contact')`, but that is less intuitive. 
- `_call_salesforce()` now gets defined three times in this codebase (twice in `api.py` and once in `bulk.py`). As a todo, might be nice to generalize and move to the `util.py` file, along with `_exception_handler()`; especially would make split services easier to implement. 
- What types of unit tests are appropriate here? Looking at the current tests, I'm not sure there's much we could add that wouldn't be dependent on API version. If any suggestions, let me know and I can either build into this PR or create a separate PR later.

### DIFF
--- a/simple_salesforce/bulk.py
+++ b/simple_salesforce/bulk.py
@@ -1,3 +1,15 @@
+""" Classes for interacting with Salesforce Bulk API """
+
+import requests
+import json
+from time import sleep
+from simple_salesforce.util import SalesforceError
+
+try:
+    from collections import OrderedDict
+except ImportError:
+    # Python < 2.7
+    from ordereddict import OrderedDict
 
 class SFBulkHandler(object):
     """ Bulk API request handler
@@ -5,13 +17,12 @@ class SFBulkHandler(object):
     This is really just a middle layer, whose sole purpose is to allow the above syntax
     """
 
-    def __init__(self, session_id, sf_instance, bulk_url, proxies=None, session=None):
+    def __init__(self, session_id, bulk_url, proxies=None, session=None):
         """Initialize the instance with the given parameters.
 
         Arguments:
 
         * session_id -- the session ID for authenticating to Salesforce
-        * sf_instance -- the domain of the instance of Salesforce to use
         * bulk_url -- API endpoint set in Salesforce instance
         * proxies -- the optional map of scheme to proxy server
         * session -- Custom requests session, created in calling code. This
@@ -72,7 +83,7 @@ class SFBulkType(object):
             'contentType': 'JSON'
         }
 
-        if operation=='upsert':
+        if operation == 'upsert':
             payload['externalIdFieldName'] = external_id_field
 
         url = self.bulk_url + 'job'
@@ -142,7 +153,7 @@ class SFBulkType(object):
 
         batch = self._add_batch(job_id=job['id'], data=data)
 
-        job_close = self._close_job(job_id=job['id'])
+        self._close_job(job_id=job['id'])
 
         batch_status = self._get_batch(job_id=batch['jobId'], batch_id=batch['id'])['state']
 
--- a/simple_salesforce/bulk.py
+++ b/simple_salesforce/bulk.py
@@ -166,32 +166,32 @@ class SFBulkType(object):
 
     # Wrappers for _bulk_operation which expose the supported Salesforce bulk operations
     def delete(self, data):
-
+        """ soft delete records """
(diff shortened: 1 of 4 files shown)


banjoh:
With regards to the comment on `_call_salesforce()`, an approach I tried is introduce a `RequestBase` class that would encapsulate the implementation of making remote requests & error handling. Any class that needs this functionality can go ahead and subclass this class. You could take a look at #149 to see how I have done it.

colemanja91:
@banjoh that's really cool! As far as I can tell it generalizes the request workflow fairly well. If that gets merged then I'll definitely plan time to pull and incorporate here too. 
@demianbrecht any thoughts?

rheineke:
@demianbrecht I've been using this branch for a few weeks, and it has worked without issue. Any plans to pull this into master?

demianbrecht:
@rheineke I'm a tad bit behind on reviews, but hoping to tackle it shortly. Thanks for the feedback! Also, thanks to @colemanja91 for putting in the work!

colemanja91:
@demianbrecht any chance you'll be able to review soon?

simple_salesforce/bulk.py
```diff
@@ -0,0 +1,338 @@
...
+        """ String together helper functions to create a complete
+        end-to-end bulk API request
+
+        Arguments:
+
+        * object_name -- SF object
+        * operation -- Bulk operation to be performed by job
+        * data -- list of dict to be passed as a batch
+        * external_id_field -- unique identifier field for upsert operations
+        * wait -- seconds to sleep between checking batch status
```
colemanja91: Small oversight, documented that parameter but didn't add. Updated. Thanks! 

Most Bulk API jobs that I've run complete in 10-30 seconds, so 5 seemed like a reasonable default to avoid waiting too long. 

I don't have much experience with `Future` objects. Also, running a SFDC Bulk API job in a multithreaded process wouldn't be my first choice. If it were a batch submit with incremental returns of individual records, then I could see the benefit. Anything I'm missing here?

colemanja91:
Dealt with latest build fail (too-many-arguments pylint error), but builds seem to be failing across the board for the same reasons... any idea what's up with that?

simple_salesforce/bulk.py
andrei-shabanski: I meant an object that would encapsulate the logic of checking a status of a job. The object would take an ID of a running job. It also would have `.result(wait = 5)` method which would run your `while` loop https://github.com/simple-salesforce/simple-salesforce/pull/143/files/97b25f64c2d9271f89f8588b85d67e61ccbe1f54#diff-0f8e08704aa000282b5b1703b2c78af3R193

See how celery, asyncio and concurrent.futures do it. They have a similar approach work. They run a job, create a `future` object and immediately return it. In this case, the current thread isn't suspended. And a user can perform some useful work while the job is processed.

davefernig:
Could @ViSovari please take a look at this PR?

ViSovari:
Pinging @chrisdonaldson on this as well to get perspective from Privacy on this.

colemanja91:
I'm not against trying to implement something more complex, but I'm not entirely sure the gains here justify any additional complexity. I can take a weekend to try and implement, to give it a fair chance.

nickcatal:
This duplicates all the exceptions. What are people's thoughts about making a new file called `exceptions.py` (or `errors.py`, or something) and putting all the exceptions used in both `api` and `bulk` in there?

colemanja91:
@nickcatal I am all for that. Do we want to try and work that in to a separate pull request first?

nickcatal:
@colemanja91 actually, after reading through things a bit since `SalesforceError` is already in `utils.py` I think we can refactor it here without creating a new PR.

maybe we move them into `utils.py` (below `SalesforceError`) and add import statements to both `api.py` and `bulk.py`? Thoughts?

PR #144, opened 2016-12-23
Pass domain along to login for govcloud domains.

MrMaksimize:
Since we use a govcloud domain, we can't login using `login.salesforce.com`

This fixes that.  

Would love feedback on what else needs changing here -- since I don't deeply know the rest of the package.

Thanks!

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -110,6 +110,7 @@ class Salesforce(object):
                 session=self.session,
                 username=username,
                 password=password,
+                instance=instance,
                 security_token=security_token,
                 sandbox=self.sandbox,
                 sf_version=self.sf_version,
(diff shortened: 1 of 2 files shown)


demianbrecht:
Thanks for the contribution @MrMaksimize and apologies for the delayed response! Without knowing all of the various combinations around SOAP-based login, I'm a little hesitant to bind the instance to the login service if set. It seems reasonable based on existing code that the `instance` parameter can be defined, but still expect the global login service to be used.

So, I think what I'd rather see here is (yet another) parameter added to `Salesforce.__init__` indicating which auth service to use. It gets a little messy because of the `sandbox` parameter, but that needs to stay at least until a major version bump.

In addition to passing it through to the login service, it'd be nice to have simple tests to verify the various permutations of parameters (specifically `sandbox` and whatever new one is added) to ensure correct behavior.

Does that make sense?

demianbrecht:
Agreed, thanks for pointing it out @shawalli. Closing, but please feel free to reopen should you feel that it's not addressed.

Issue #148, opened 2017-01-28
Ability to pass request timeouts to Salesforce http operations


banjoh:
In an automated environment, requests to Saleforce not timing can cause really bad service interruptions. I've had cases where a service has "gone silent" for a few days. At the moment most of the API calls have no mechanism to pass in the timeout parameter to `requests`. I've had blocked request when trying to login and when performing operations related to `sobjects`.

Generally we should be able to pass down parameters to `requests`. This issue is also similar to #81

ddow:
This request is full of win.

ivanpricewaycom:
Hello, we are interested in this merge, is there a timeline / estimate for a release ?

cheers

JackaDu:
Me, too. The salesforce connection has been silent for 7 hours

alexprice1:
+1

pdr:
In light of recent Salesforce issues, it seems like this is getting more and more critical.

chingc:
+1

nmassman:
So you can actually set a timeout for every call of a particular requests session by attaching your own HTTP adapter. This is the solution I've used in my own usage of simple_salesforce as well as for other modules that let you provide your own requests.Session() object.

```
class TimeoutAdapter(requests.adapters.HTTPAdapter):
    def send(self, *args, **kwargs):
        kwargs['timeout'] = 30
        return super().send(*args, **kwargs)

sfSession = requests.Session()
sfSession.mount("https://", TimeoutAdapter())
sfSession.mount("http://", TimeoutAdapter())
```
Then pass `session=sfSession` when you instantiate your Salesforce object.

There is actually a lot of powerful things you can do with overriding parts of the default HTTPAdaptor in requests.
https://requests.readthedocs.io/en/master/user/advanced/#transport-adapters

PR #149, opened 2017-01-31
Support python requests keyword arguments when making remote calls

banjoh:
Since this library is heavily dependent on python requests for making remote calls, it is necessary to pass down parameters such as timeout and header to requests methods.

This PR is related to #148


simple_salesforce/api.py
```diff
@@ -774,22 +803,12 @@ def __init__(self, username, password, security_token, sandbox=False,
 
 def _exception_handler(result, name=""):
```
banjoh: `def _exception_handler` could be removed totally. I left it here not to brake any other forks using it. It's private, but I assume using a proper deprecation process would be best.

mattbriancon:
@demianbrecht @nickcatal Not sure who to ping but is there any chance we could get this merged soon? Happy to help make tweaks to get it mergeable.

peppyCola:
Going to bring this up again; not sure how maintained this library is anymore but would be great to have this feature merged to avoid needing to patch.

Issue #109, opened 2016-06-03
Salesforce.restful call is broken for POST


dursk:
Specifying the method for `.restful` was recently added, however, the way the `params` end up getting passed down to the `requests` library needs to change depending on if it's a `GET` or `POST` request.

Currently, if I call `Salesforce.restful(url, {'foo': 1}, method='POST')`, it results in the payload getting passed in as query parameters rather than the body.

demianbrecht:
Thanks for the report @dursk! Looks like the problem is that `restful` doesn't actually allow for a body. The second parameter (`params`) is intended to be passed through query params.

I think the fix here would be to add an optional `data` param to `restful` and then propagate that down through `_call_salesforce`s `kwargs`. I'm not sure when I'll be able to get to this, feel free to submit a patch if you get time before I do.

PR #155, opened 2017-03-29
Allows restful to pass http body data along with a http request.

hewe:
Allows restful to accept and pass http body data down into _call_salesforce(..). This resolves bug #109


tedmiston:
Closing this as superseded by https://github.com/simple-salesforce/simple-salesforce/pull/242

PR #156, opened 2017-03-30
Fix unicode handling in SalesforceError

Daenyth:
I confirmed with tox that the test fails without the change to util.py and passes with it.

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -798,7 +798,7 @@ class SalesforceMoreThanOneRecord(SalesforceError):
     The value returned when an external ID exists in more than one record. The
     response body contains the list of matching records.
     """
-    message = u"More than one record for {url}. Response content: {content}"
+    message = u"More than one record for {0.url}. Response content: {0.content}"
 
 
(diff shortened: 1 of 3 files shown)


Daenyth:
I also recommend using the [`future`](http://python-future.org/index.html) library for compatibility with python2+3 - it's a huge time saver and has a [convenient decorator](http://python-future.org/compatible_idioms.html#custom-str-methods) for this case.

Daenyth:
`tox -e static` passes locally for me - anyone know what's up with that?

Daenyth:
cc @demianbrecht

demianbrecht:
@Daenyth do you have global pylint config that may be overriding the local one?

Daenyth:
Not that I'm aware of. Shouldn't the `tox` build be taking care of that anyway?

demianbrecht:
Yeah, it should be. I'll try to take a look at what's going on today. Thanks!

Daenyth:
I probably won't be coming back to this as I'm not working in python at the moment - anyone else feel free to take up this branch and take it the rest of the way

PR #162, opened 2017-05-10
Added license to the manifest

mariusvniekerk:
In order to easily package this library on conda forge we should include the license file in the source tarball.


nickcatal:
Thanks for this

PR #165, opened 2017-05-30
Fix lint errors

lambacck:
#164 tests are failing due to lint errors. This should fix the lint errors. PyLint is not pinned to a version so as new versions are released there could be random static analysis failures base on new linters. These look like they were introduced in [PyLint 1.7](https://pylint.readthedocs.io/en/latest/whatsnew/1.7.html).

### DIFF
--- a/simple_salesforce/util.py
+++ b/simple_salesforce/util.py
@@ -16,7 +16,7 @@ def getUniqueElementValueFromXmlString(xmlString, elementName):
     xmlStringAsDom = xml.dom.minidom.parseString(xmlString)
     elementsByName = xmlStringAsDom.getElementsByTagName(elementName)
     elementValue = None
-    if len(elementsByName) > 0:
+    if elementsByName:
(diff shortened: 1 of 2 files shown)


simple_salesforce/util.py
```diff
@@ -16,7 +16,7 @@ def getUniqueElementValueFromXmlString(xmlString, elementName):
     xmlStringAsDom = xml.dom.minidom.parseString(xmlString)
     elementsByName = xmlStringAsDom.getElementsByTagName(elementName)
     elementValue = None
-    if len(elementsByName) > 0:
+    if elementsByName:
```
lambacck: This returns a `NodeList` which is a subclass of `list` and is therefore safe to do a normal empty test as above.

andscoop:
This looks resolved by either changes in pylint version or @nickcatal 's commit
https://github.com/simple-salesforce/simple-salesforce/commit/e42c1c47eca206b03685a89460b52d87c6b69408

Are we good to close?

PR #166, opened 2017-05-31
Use sudo: false

mattbriancon:


mattbriancon:
pypy tests breaking because cryptography was updated to require pypy>=5.4.

Happy to require `pypy-5.4.1` in `.travis.yml` (preferred) or pin the `requests[security]` requirement back to a known working version.

Issue #163, opened 2017-05-25
can we send email via salesforce?


llcao:
Hi 
I am new to simple-salesforce but I feel it very cool!

My question is: can we use them in simple-salesforce?

I saw salesforce provides the API of sending email
https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_calls_sendemail.htm

skamensky:
@llcao , the simple-salesforce api doesn't have a wrapper around this sendEmail endpoint.

@demianbrecht , is this within the scope of the simple-salesforce project?

llcao:
Thanks @skamensky 

I can help to implement a wrapper for the sendEmail endpoint if you and @demianbrecht are interested.  Should it be a new function in Salesforce class?

afs2015:
@llcao this would be functionality that I would be interested as well. However, I can understand if it is outside of the scope of this project.

skamensky:
I haven't had any personal use cases for this but I'd be glad to help implement and maintain it as my first go at simple-salesforce.

 We would need some direction from some of the core maintainers as to whether or not this is a feature we should implement, what conventions there are for feature development, testing, and CI.

llcao:
Update: I did a little research. it seems to me that we can use 
`/services/data/v32.0/actions/standard/emailSimple`
to send email.
I think it could be implemented by a customized function for SFType Action. 
Will work on it in the next few days. 

Suggestions are welcome.

skamensky:
I think it makes sense to add it as a method of the Salesforce class since sending emails isn't tied to a specific object and is a more general action like other methods in the Salesforce class like "apexecute," "search," and "setPassword."

Another suggestion. This is the second place we're using SOAP. The first place is in login.py.

We should create a utility function in utility.py to handle SOAP data string creation and other SOAP functions. We can do a refactoring of login.py at a different date to use that utility.

llcao:
Here is my implementation. I'd submit a pull request if people are interested. 
https://github.com/llcao/simple-salesforce
It works with my salesforce standard account.

Basically I added a new class SFemailSimple and add it to def __getattr__(self, name) as  
``` 
def __getattr__(self, name):
    if name.startswith('__'):
            return super(Salesforce, self).__getattr__(name)

    if name == 'Email':  return SFemailSimple(...)

   return SFType(...)
```

Several notes:

[...]

PR #170, opened 2017-06-15
Send simple Email via simple-salesforce

llcao:
this is a  function motivated by issue #163. 

Basically I added a new class SFAction and add it to Salesforce.getattr(self, name) as the following:

```
def __getattr__(self, name):
    if name.startswith('__'):
            return super(Salesforce, self).__getattr__(name)

    if name == 'Action':  return SFAction(self.session_id, self.sf_instance,
                            proxies=self.proxies, session=self.session)

   return SFType(...)
```
and the SFAction class implements emailSimple() and hopefully also other actions in the future.

Several notes:

- I use REST API instead of SOAP based on this [doc](https://resources.docs.salesforce.com/sfdc/pdf/api_action.pdf). I don't know whether SOAP is a better choice. 

- simpleEmail exists only after version 32.0. With DEFAULT_API_VERSION = '29.0', it will report some error like "resource does not exist".

I also tried with my sandbox account and got correct status_code, but I just could not receive the email. Haven't figured out the reason, just hoping the problem lies in my sandbox setting instead of the code.

Thanks for the suggestions. More comments are welcome.

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -201,9 +201,9 @@ class Salesforce(object):
         if name.startswith('__'):
             return super(Salesforce, self).__getattr__(name)
 
-        if name == 'Email':
-            return SFemailSimple(self.session_id, self.sf_instance,
-                                 proxies=self.proxies, session=self.session)
+        if name == 'Action':
+            return SFAction(self.session_id, self.sf_instance,
+                            proxies=self.proxies, session=self.session)
 
         return SFType(
             name, self.session_id, self.sf_instance, sf_version=self.sf_version,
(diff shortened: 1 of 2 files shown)


skamensky:
Why do all checks fail?

llcao:
I could not figure out the reason. @skamensky could you give some hints? (I had very limited experience with travis-ci)

skamensky:
@llcao me neither. If you click into the details you see that for all the non-allowed failures, you have an error like

`E: 41,11: Unexpected keyword argument 'object_name' in constructor call (unexpected-keyword-arg)
`
https://travis-ci.org/simple-salesforce/simple-salesforce/jobs/243313606 gives a traceback of 

```
Traceback (most recent call last):
  File "<string>", line 3, in wrapper
  File "/home/travis/build/simple-salesforce/simple-salesforce/simple_salesforce/tests/test_api.py", line 58, in test_version
    emailSimple = _create_sf_emailSimple()
  File "/home/travis/build/simple-salesforce/simple-salesforce/simple_salesforce/tests/test_api.py", line 45, in _create_sf_emailSimple
    session=requests.Session()
TypeError: __init__() got an unexpected keyword argument 'object_name'
```

afs2015:
@llcao: Do you want help with debugging this? I'm using workforce rules and email alerts to send emails out right now but would prefer to use Salesforce's API if possible.

llcao:
Thank you @afs2015 for reminding me. And sorry for my delay. I wrongly thought I have fixed the problem and totally forgot this thread. Now I am submitting a new pull request and hopefully it will remove all the bug. 

I am still not familiar with Travis Cl. Will check again after it finishes.  Comments and suggestions are welcome.

llcao:
Could someone help? I met an error from Travis Cl. 
```
ERROR:   static: commands failed
```
but do not understand why it happens and how to fix it.

afs2015:
Awesome work @llcao!

PR #177, opened 2017-08-08
Handle 204 in set password

fameoflight:
@nickcatal

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -242,11 +242,15 @@ class Salesforce(object):
                                          'User',
                                          result.status_code,
                                          result.content)
-        json_result = result.json(object_pairs_hook=OrderedDict)
-        if len(json_result) == 0:
-            return None
 
-        return json_result
+        if result.status_code != 204:
+            json_result = result.json(object_pairs_hook=OrderedDict)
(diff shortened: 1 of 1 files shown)


nickcatal:
@fameoflight can you explain what issue this PR is fixing?

fameoflight:
Handling 204 correctly. 

204 means no response so don't try to read the content if the status type is 204.

nickcatal:
Got it.

To be consistent with how `.delete()` works, can you have it return the status code of the request instead of `None`?

Then I can merge this in

andscoop:
@nickcatal Should we push for code consolidation? Directly above this change we are already checking for `status_code != 204`.

@fameoflight Is there any advantage to doing a check for a 204 and returning None vs just checking the JSON body and returning None if len is zero?

andscoop:
Looking at docs in code comment, I think we should throw an error `status_code != 204` and `return None` `if status_code == 204`. That would be an easy fix for now.

In the long term it would be nice to see all `status_code` logic moved into it's own method off of `_call_salesforce` similar to how we handle >=300 errors now.

I do not believe this code get's us close enough to a final solution to warrant a merge.

PR #187, opened 2017-10-12
Deprecated sandbox flag in favor of domain flag

shawalli:
This is useful for cases where the login domain is a custom domain, meaning that
it is neither 'login' nor 'test'. Default is 'login', however, so the default behavior is unchanged. Updated current api appropriately (along with non-fatal deprecation warning), while allowing SalesforceApi class to stay the same.

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -104,8 +104,8 @@ class Salesforce(object):
             if kwargs['sandbox'] is True:
                 login_domain = 'test'
 
-        # Determine if the user passed in the optional version and/or login_domain
-        # kwargs
+        # Determine if the user passed in the optional version and/or
+        # login_domain kwargs
         self.sf_version = version
         self.login_domain = login_domain
         self.session = session or requests.Session()
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -59,9 +59,8 @@ class Salesforce(object):
     def __init__(
             self, username=None, password=None, security_token=None,
             session_id=None, instance=None, instance_url=None,
-            organizationId=None, version=DEFAULT_API_VERSION,
-            proxies=None, session=None, client_id=None, login_domain='login',
-            **kwargs):
+            organizationId=None, sandbox=None, version=DEFAULT_API_VERSION,
+            proxies=None, session=None, client_id=None, domain=None):
         """Initialize the instance with the given parameters.
 
         Available kwargs
@@ -71,9 +70,11 @@ class Salesforce(object):
         * username -- the Salesforce username to use for authentication
         * password -- the password for the username
         * security_token -- the security token for the username
-        * sandbox -- DEPRECATED: Use login_domain instead.
-        * login_domain -- The domain to using for logging in. Use common login
-                          domains, such as 'login' or 'test', or named domain.
+        * sandbox -- DEPRECATED: Use domain instead.
+        * domain -- The domain to using for connecting to Salesforce. Use
+                    common domains, such as 'login' or 'test', or
+                    Salesforce My domain. If not used, will default to
+                    'login'.
 
         Direct Session and Instance Access:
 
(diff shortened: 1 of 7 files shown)


skamensky:
Looks beautiful. Thanks for sharing!

shawalli:
@skamensky sure! There are a number of ways to add this functionality. However, I chose this route because it did not increase complexity by leaving the "sandbox" flag and adding another flag or by changing the "sandbox" flag's expected type.

simple_salesforce/api.py
```diff
@@ -59,8 +59,9 @@ class Salesforce(object):
     def __init__(
             self, username=None, password=None, security_token=None,
             session_id=None, instance=None, instance_url=None,
-            organizationId=None, sandbox=False, version=DEFAULT_API_VERSION,
-            proxies=None, session=None, client_id=None):
+            organizationId=None, version=DEFAULT_API_VERSION,
+            proxies=None, session=None, client_id=None, login_domain='login',
+            **kwargs):
```
demianbrecht: Can you remove `kwargs` and replace it with `sandbox` for the time being, keeping the same deprecation warning in place?

simple_salesforce/login.py
```diff
@@ -16,13 +16,15 @@
...
 
 
 # pylint: disable=invalid-name,too-many-arguments,too-many-locals
 def SalesforceLogin(
         username=None, password=None, security_token=None,
-        organizationId=None, sandbox=False, sf_version=DEFAULT_API_VERSION,
-        proxies=None, session=None, client_id=None):
+        organizationId=None, sf_version=DEFAULT_API_VERSION,
+        proxies=None, session=None, client_id=None, login_domain='login',
+        **kwargs):
```
demianbrecht: Same thing here with the `kwargs`

simple_salesforce/login.py
```diff
@@ -34,19 +36,28 @@ def SalesforceLogin(
...
                  exposed by simple_salesforce.
     * client_id -- the ID of this client
+    * login_domain -- The domain to using for logging in. Use common login
+                      domains, such as 'login' or 'test', or named domain.
     """
-
-    soap_url = 'https://{domain}.salesforce.com/services/Soap/u/{sf_version}'
-    domain = 'test' if sandbox else 'login'
+    if 'sandbox' in kwargs.keys():
+        warnings.warn('\'sandbox\' argument is deprecated. Use '
```
demianbrecht: Super nitpick here, but rather than escaping, can you alternate single and double quotes? Whichever you prefer. Not merge blocking, I only find it more aesthetically pleasing when reading through code.

simple_salesforce/login.py
```diff
@@ -34,19 +36,28 @@ def SalesforceLogin(
...
+                        '\'login_domain\' instead. Overriding login_domain '
+                        'with sandbox flag.',
+                        DeprecationWarning)
+        login_domain = 'login'
+        if kwargs['sandbox'] is True:
```
demianbrecht: This isn't strictly backwards compatible. A lot that isn't `True` would have previously evaluated to true due to truthiness tests ('asd', [1], 1, etc). Can you drop the `is True` here?

simple_salesforce/login.py
```diff
@@ -34,19 +36,28 @@ def SalesforceLogin(
...
                  enables the use of requets Session features not otherwise
                  exposed by simple_salesforce.
     * client_id -- the ID of this client
+    * login_domain -- The domain to using for logging in. Use common login
+                      domains, such as 'login' or 'test', or named domain.
     """
-
-    soap_url = 'https://{domain}.salesforce.com/services/Soap/u/{sf_version}'
-    domain = 'test' if sandbox else 'login'
+    if 'sandbox' in kwargs.keys():
```
demianbrecht: Perhaps add an assertion here ensuring both parameters aren't being used? Given the logic below, that'd lead to unexpected results.

simple_salesforce/tests/test_login.py
```diff
@@ -75,6 +75,6 @@ def test_failure(self):
                 username='myemail@example.com.sandbox',
```
demianbrecht: Can you add tests here for to ensure:

1. The correct domain is used when specified as a parameter?
2. Both parameters being used will result in some kind of an error?

simple_salesforce/login.py
shawalli: The assertion wouldn't really work if `sandbox` is expected to be True or False (translating to 'login' or 'test'). I can either default to keyword to None, and then assert treat True/False as being used. Otherwise, we'd need to come up with some other assertion logic.

shawalli: I went ahead and set sandbox=None and domain=None, and then updated the logic appropriately. This is the only way I could think to fully allow both arguments but still regulate the expected behavior.

ptn24:
What is the status of this? Any chance it can be merged and shipped?

tmhammer:
@demianbrecht, Any updates on this?

ptn24:
Ping

shawalli:
@demianbrecht  Is there a reason that this isn't being incorporated? Status checks are passing

demianbrecht:
Hey @shawalli, thanks for the ping, merged. I've taken a seat _way_ in the back for the time being. Thanks again for the submission. It won't be available from PyPI until the next release. @andscoop or @tedmiston would you mind doing a release whenever you have a chance so this can get out?

tedmiston:
Sure, @andscoop and I can cut the release next week.  I'm not super familiar with the versioning scheme here.  Are you thinking v0.74.0 next?  Also, can you give us maintainer access on PyPI / Warehouse?

demianbrecht:
@tedmiston I've added you as a maintainer on Warehouse. @andscoop can you ping me with your Warehouse user name if you'd like to be added as well?

@tedmiston I was kinda just trudging along with the previous versioning. I think we wanted to hit 100% coverage before releasing 1.0.0, but one could argue that the package is mature enough to be released as a 1.0.0. Up to you guys really, I don't really have a strong opinion on it (of course, that wouldn't be the same going forward, I'd like to see something close to semver).

tedmiston:
Sounds good

andscoop:
@demianbrecht All setup on pypi - can you add me (andscoop)?

PR #188, opened 2017-10-16
Added Header

BenPru:
Added header to json result. This allows you to read the current API usage and all other headers.

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -194,6 +194,7 @@ class Salesforce(object):
         json_result = result.json(object_pairs_hook=OrderedDict)
         if len(json_result) == 0:
             return None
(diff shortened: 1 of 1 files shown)


demianbrecht:
I'm not sure I'm a fan of this approach because of the chance of overwriting object attributes. I understand the problem and put some thought into it a while ago. What I'd prefer to see instead is a deserializer added to the `Salesforce` class. When instantiating, you could pass in `json.loads` (or some module function that mirrors current functionality), or `None` (defaulting to the json method). On API calls, the specified deserializer would be called. If `None`, the raw response object would be returned, which would provide the functionality we're after here.

The only question that I really have is if instance-level attributes are good enough, or if folks also want this to be added consistently across the API calls (I'd prefer the first but understand the inconvenience).

PR #190, opened 2017-10-16
Typo in Readme

catalandres:
Intro text to lists of delete/updated records did not match code.


demianbrecht:
Thanks for the catch!

PR #192, opened 2017-10-20
Add method for getting organization limits

leonardoam:
I made this for myself and I think it may be useful for someone.

Add a simple method for getting organization limits as described in
https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_limits.htm

This is my first time PR in a open repo, so please be patient if a missed something stupid.

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -339,6 +339,18 @@ class Salesforce(object):
         search_string = u'FIND {{{search_string}}}'.format(search_string=search)
         return self.search(search_string)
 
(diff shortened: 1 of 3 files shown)


demianbrecht:
Looks useful to me, thanks! I'd like to see tests added for it first prior to merging though.

leonardoam:
I've added a test. It's a very simple request, so I think it may be enough.
Please let me know your thoughts.

andscoop:
@leonardoam Thank you for the work on this, especially the additional tests.

PR #207, opened 2017-12-27
#146 simplify errorhandling and handle non 200 success codes

### LINKED ISSUE #146: restful() should handle non-200 success codes
There are endpoints in the Salesforce API that result in non `200` response codes (i.e. 201 from [upserts](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_upsert.htm)). Currently, `restful()` will result in an exception being raised if the response code is [not strictly `200`](https://github.com/simple-salesforce/simple-salesforce/blob/ef0a9487a6cd4854466d00969c32da000d34392a/simple_salesforce/api.py#L271). This should be updated to accept anything outside of obvious error code ranges (`4xx` and `5xx`).

andscoop:
- Fixes #146 
- Removes redundant status code checks
- refactor _call_salesforce to accept name so that error handling/ status code checks can all happen centrally
- brings up code quality to 10/10

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -185,12 +185,8 @@ class Salesforce(object):
         """Describes all available objects
         """
         url = self.base_url + "sobjects"
-        result = self._call_salesforce('GET', url)
-        if result.status_code != 200:
-            raise SalesforceGeneralError(url,
-                                         'describe',
-                                         result.status_code,
-                                         result.content)
+        result = self._call_salesforce('GET', url, name='describe')
+
         json_result = result.json(object_pairs_hook=OrderedDict)
         if len(json_result) == 0:
             return None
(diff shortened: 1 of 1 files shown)


simple_salesforce/api.py
```diff
@@ -431,18 +414,19 @@ def apexecute(self, action, method='GET', data=None, **kwargs):
...
+            return result.json()
+        # pylint: disable=broad-except
+        except Exception:
+            return result.text
 
-    def _call_salesforce(self, method, url, **kwargs):
+    def _call_salesforce(self, method, url, name="", **kwargs):
         """Utility method for performing HTTP call to Salesforce.
 
         Returns a `requests.result` object.
```
sigmavirus24: What is `requests.result`. Do you mean `requests.Response`?

PR #211, opened 2018-01-07
multi batch jobs

mbr84:
When using bulk api, split records into batches containing <= 10000 records, and associate these with a single job

### DIFF
--- a/simple_salesforce/bulk.py
+++ b/simple_salesforce/bulk.py
@@ -195,17 +195,19 @@ class SFBulkType(object):
         def worker():
             """ Add batches and wait for network responses (I/O) concurrently
             with worker threads. (there may be a lot of batches)
-
+            
             Arguments: 
             (Workers pull their args from the queue)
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -18,23 +18,15 @@
 
 here = os.path.abspath(os.path.dirname(__file__))
 
-about = {}
-with open(os.path.join(here, 'simple_salesforce', '__version__.py'), 'r') as f:
-    exec(f.read(), about)
-
 
 setup(
-    name=about['__title__'],
```
andscoop: Remove your fork specific details

mbr84: sorry my fault. I have to get a version of this running on a project for work. The latest commit was meant for another repo. Should be fixed now

mbr84:
@andscoop in case it makes a difference, it can do this now with one call...
screen shot 2018-01-10 at 7 38 35 am

jankatins:
There was just something similar merged, this can probably be closed

jon-wobken:
Solved by #392

Issue #213, opened 2018-01-18
SalesforceGeneralError parameters out of order in Salesforce.restful


sigmavirus24:
`Salesforce.restful` seems to raise a `SalesforceGeneralError` [here](https://github.com/simple-salesforce/simple-salesforce/blob/dda2708f13e815566f87f182601d23717797632b/simple_salesforce/api.py#L293) but the ordering of the parameters looks wrong. I think it should be 

```py
SalesforceGeneralError(
    url,
    response.status_code,
    path,
    response.content,
)
```

but right now status code and path are swapped. See [SalesforceError](https://github.com/simple-salesforce/simple-salesforce/blob/dda2708f13e815566f87f182601d23717797632b/simple_salesforce/exceptions.py#L9)'s `__init__`.

PR #214, opened 2018-01-18
Update usage of SalesforceGeneralError in restful

sigmavirus24:
Use keyword arguments to pass the values to SalesforceGeneralError
so they're stored appropriately on the exception.

Closes gh-213

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -290,10 +290,10 @@ class Salesforce(object):
         url = self.base_url + path
         result = self._call_salesforce(method, url, params=params)
         if result.status_code != 200:
-            raise SalesforceGeneralError(url,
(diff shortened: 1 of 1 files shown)


sigmavirus24:
I don't believe https://travis-ci.org/simple-salesforce/simple-salesforce/jobs/330445312#L472 is related to my changes and I'm not confident I should fix that up in this PR.

andscoop:
@sigmavirus24 Thanks for doing this - but this change goes away with #146 which has been approved and will be merged shortly. Reopen if there is any objection.

sigmavirus24:
@andscoop I opened a bug for some of the problems that change has illustrated for me. (I've only just started using `simple_salesforce`.)

Issue #212, opened 2018-01-07
Pull Request test


mbr84:
Hey all, I'm having a pull request blocked by a linter error in file I'm not working on, api.py. 
```R:424, 4: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)```
The error is on line 424, and when I look it up, it seems to have been identified as a bug and removed from pylint. Any guidance what I should do to move forward?

andscoop:
Looking at the git blame linked below it appears that trying to get fancy and return directly from `.json()` and `.text` in the method/attr resulted in this pylint error. I am not entirely sure why I was able to pass build test though. Either way,  I guess if I broke it I'm going to have to buy it. I will put a fix out shortly, which should resolve your issues.

https://github.com/simple-salesforce/simple-salesforce/blame/master/simple_salesforce/api.py#L424

mbr84:
That fixed it. hooray!

PR #216, opened 2018-01-24
fix inconsistent-return-statement linter error

andscoop:
This fix should allow us to close out #212


tedmiston:
LGTM

I see the idea the linter rule was going for in the linked issue, but also that this doesn't change things functionally.  I think it'd be worth our time to tighten the specific exceptions that are caught here as well in the future.

PR #196, opened 2017-11-10
Allow for POST support in restful method

dursk:
Fix for https://github.com/simple-salesforce/simple-salesforce/issues/109

PR #224, opened 2018-02-09
Allow transmitting a body in restful().

filmor:


filmor:
Same as #196, sorry for the noise.

PR #225, opened 2018-02-11
rm python 3.3 from travis ci

andscoop:
Latest version of pylint drops support for python 3.3. Updating .travis.yml to allow failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ script: tox
 matrix:
   fast_finish: true
   allow_failures:
+    # pylint 1.8.2 dropped support for python 3.3
+    - python: "3.3"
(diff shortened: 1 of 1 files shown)


tedmiston:
So is the idea here to deprecate support for 3.3 or just disable the linter on it?

I see that simple-salesforce supports 2.6, 3.2, 3.3 today but these versions aren't popularly used in production much.  It might be worth considering deprecating those in the next major release.

andscoop:
Agreed. Issue here was specifically pylint not supporting 3.3 any longer which was causing our travis-ci builds to break.

tedmiston:
I saw in https://github.com/exercism/python/issues/1157 that Python 3.3 was EOL'd in September 2017 as well.  All the more reason to drop support there.

https://www.python.org/dev/peps/pep-0398/#x-end-of-life

PR #233, opened 2018-03-13
Fix bulk upsert query example

### LINKED ISSUE #232: Bulk upsert example incorrect
In the Readme, the bulk upsert data example is as follows:

```python
data = [
  {'Id': '0000000000AAAAA', 'Email': 'examplenew2@example.com'},
  {'Id': '', 'Email': 'foo@foo.com'}
]
```
based on my testing and usage of this query, dictionaries that had `'Id': ''` were not actually inserted. Only by removing the `Id` key was I able to get it to work. 

I'm happy to make the PR to update this unless there's something I'm missing.

tayloramurphy:
The bulk upsert query does not work properly if the `Id` is an empty string. I was able to get it to work by removing the `Id` key altogether. This PR updates the example and improves the styling on the other bulk query examples.

Fixes #232

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -280,9 +280,7 @@ Delete records (soft deletion):
 
 .. code-block:: python
 
-    data = [
-          {'Id': '0000000000AAAAA'}
-        ]
+    data = [{'Id': '0000000000AAAAA'}]
 
     sf.bulk.Contact.delete(data)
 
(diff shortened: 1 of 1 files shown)


andscoop:
Thanks for taking the time to correct the documentation.

PR #254, opened 2018-05-05
Improve PyPI releases done via Travis CI

andscoop:
This PR

- will skip cleanup of build artifacts
- will only release to PyPi when a new release is cut for master branch
- limits pypi releases to only being performed for python3.6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,12 +30,14 @@ deploy:
   # Allow travis-ci to deploy new tags to PyPI
   provider: pypi
   user: simple-salesforce
+  skip_cleanup: true
   password:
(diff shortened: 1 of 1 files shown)


andscoop:
A description from the issue I was going to create, before finding an easy resolution and creating this pr

>Somewhere between [release v0.74.1](https://travis-ci.org/simple-salesforce/simple-salesforce/builds/375179419) and [release v.73.0](https://travis-ci.org/simple-salesforce/simple-salesforce/builds/249219996) the `.travis.yml` has become more strict. This has given the appearance of hard failures during a PyPi release via CI. Looking at the links above and digging into the logs you can see that `HTTPError: 400 Client Error: File already exists.` is happening in both builds, something in v.0.73.0 has just allowed for this failure.

tedmiston:
@andscoop This is the fix we were talking about Friday night, right?

Looking at it again, I think I may have misunderstood the [`skip_cleanup`](https://docs.travis-ci.com/user/deployment#Uploading-Files-and-skip_cleanup) option:

> When deploying files to a provider, prevent Travis CI from resetting your working directory and deleting all changes made during the build ( git stash --all) by adding skip_cleanup to your .travis.yml:

Actually, now I think we *would* want it to stash / reset any local changes that are not in the repo before pushing the build to PyPI.  I haven't looked to see if the wheel is built in this directory or another directory, but that's the only asset I can think of that we have here (right?)

Is this something we can test pre-merge?

demianbrecht:
Why would we want releases to be limited to 3.6?

PR #258, opened 2018-06-19
fix: exception thrown when salesforce returns 204

DovAmir:
exception thrown when salesforce returns 204.
this happens when doing PATCH request in record field because reponse is empty string and cant be 
serialized
result.json(object_pairs_hook=OrderedDict)

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -303,7 +303,10 @@ class Salesforce(object):
         url = self.base_url + path
         result = self._call_salesforce(method, url, name=path, params=params,
                                        **kwargs)
-
+        
+        if result.status_code == 204:
+            return None
+        
         json_result = result.json(object_pairs_hook=OrderedDict)
         if len(json_result) == 0:
             return None


metadaddy:
I see this same problem (exception on 204 response). My solution is very similar:

```
    # Generic Rest Function
    def restful(self, path, params=None, method='GET', **kwargs):
        """Allows you to make a direct REST call if you know the path

        Arguments:

        * path: The path of the request
            Example: sobjects/User/ABC123/password'
        * params: dict of parameters to pass to the path
        * method: HTTP request method, default GET
        * other arguments supported by requests.request (e.g. json, timeout)
        """

        url = self.base_url + path
        result = self._call_salesforce(method, url, name=path, params=params,
                                       **kwargs)
        if len(result.content) > 0:
            json_result = result.json(object_pairs_hook=OrderedDict)
        else:
            return None

        return json_result
```

@DovAmir Either way, this needs to be integrated as the current behavior breaks DELETE and PATCH requests.

thommor:
Thanks @metadaddy. Let me work out why the Travis builds are failing and I’ll merge this in ASAP.

metadaddy:
@thommor The Travis failures look like `pylint` issues to me. Not sure what it's complaining about.

PR #264, opened 2018-08-04
Supporting non-json REST endpoints

nisseknudsen:
When entering a rest URL in the `restful` function, it automatically returns the parsed json response. For downloading files from Salesforce, e.g., _ContentDocumentVersion_, the response is of `Content-Type: application/octet-stream` (interestingly, the real value in the response header from Salesforce is _application/octetstream_, in contrast to their documentation).

The changed `restful` function now reads the _Content-Type_ value and parses the json for `Content-Type: application/json` and returns an unparsed response for all other _Content-Type_ values.

When querying ContentDocumentVersion, the response from Salesforce is a REST path which includes the `/services/data/vXX.X/` prefix. To provide a fluent usage of simple-salesforce with other applications, the `restful` function now accepts the prefix, and removes it by using a regex.

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -289,7 +289,9 @@ class Salesforce(object):
 
     # Generic Rest Function
     def restful(self, path, params=None, method='GET', **kwargs):
-        """Allows you to make a direct REST call if you know the path
+        """Allows you to make a direct REST call if you know the path.
+        If `Content-Type: application/json` in response, function will return a JSON-parsed OrderedDict.
+        For all other cases the raw content body is returned, e.g., binary data.
 
         Arguments:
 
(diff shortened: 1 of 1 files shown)


andscoop:
I would like to see some more feedback from the community before doing anything with this one. I'm concerned that we are now returning two types and that may break downstream code. 

Personally i'm starting to lean towards having most of these API calls return the `requests.response` object by default and then adding a flag on `Salesforce` class to make a best effort to return json.

I would also need to see tests written for this before I'd be willing to merge.

simple_salesforce/api.py
```diff
@@ -299,16 +302,16 @@ def restful(self, path, params=None, method='GET', **kwargs):
         * method: HTTP request method, default GET
         * other arguments supported by requests.request (e.g. json, timeout)
         """
-
-        url = self.base_url + path
+        base_regex = r'\/*services\/data\/v[0-9]{2}.[0-9]{1,}\/'
```
haroldo-bonette, line 305: this is something that you really should consider to merge since resources URIs include "services/data/vXX.X" in it, so if any tries to search any resource using the URI returned in a previous call it will not work as it now assumes resources URIs dont include the services/data/vXX.X part.

jon-wobken: Not opposed to merging this in just need to review it.

PR #265, opened 2018-08-23
Allow additional headers for Salesforce._call_salesforce

jgillmanjr:
Relatively simple fix to allow the `_call_salesforce()` method of the `Salesforce` class to accept additional headers, similar to how the method can accept them for the `SFType` Class.

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -471,8 +471,12 @@ class Salesforce(object):
 
         Returns a `requests.result` object.
         """
+        headers = self.headers.copy()
+        additional_headers = kwargs.pop('headers', dict())
+        headers.update(additional_headers)
+
         result = self.session.request(
-            method, url, headers=self.headers, **kwargs)
+            method, url, headers=headers, **kwargs)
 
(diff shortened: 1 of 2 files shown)


andscoop:
Thank you for your contribution, normally I would push back to require testing, but after reviewing the code I tend to agree that this is a simple fix that I am comfortable pushing through. Ty.

andscoop:
Thank you for your contribution, normally I would push back to require testing, but after reviewing the code I tend to agree that this is a simple fix that we can push without testing.

Are you able to fix the linting errors before I merge?

jgillmanjr:
I actually forgot about this PR.

Yeah, let me try and get those cleaned up.

andscoop:
@jgillmanjr Actually you may be able to simply rebase onto master to resolve this issue. I locked down pylint a few months back to take care of the particular linting error you were receiving.

jgillmanjr:
Yeah. I derped up by not rebasing before doing 4683d2d

Now I get to play the figure out how to clean this up game.

Sorry about that :/

PR #293, opened 2019-02-15
add ability to pass the file parameter to POST request for creating an Attachment

nishantd:
and also remove the Content-Type header.

This is required in order to upload an attachment using a multipart upload.

To upload an attachment /tmp/image.png, we need to make the following request:

```
files = {"Body":('image.png',open('/tmp/image.png', 'rb'), 'image/png'),
            "entity_document":(None, json.dumps({"Name":"image.png"}), "application/json")}

r = session.request('POST', 'https://my.salesforce.instance/services/data/v38.0/sobjects/Attachment/', 
                      headers={"X-header-1":"bearer token"}, 
                      files=files
                     )
```

Currently this is not possible because all create requests do a POST with the data parameter, and also the 'Content-Type: application/json' header.

I allow a new parameter 'file', and also allow 'Content-Type' header to not be passed. When the 'file' parameter is passed to session.request, it creates its own multi-part header, with the correct separator. I.e.
```
Content-Type: multipart/form-data; boundary=6fcba61d5e6b467ba8d5983c120cfabf
```

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -34,6 +34,7 @@ except ImportError:
 #pylint: disable=invalid-name
 logger = logging.getLogger(__name__)
 
+
 def _warn_request_deprecation():
     """Deprecation for (Salesforce/SFType).request attribute"""
     warnings.warn(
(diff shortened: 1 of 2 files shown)


nishantd:
Will add a test for this. Just created the PR to get any feedback.

nishantd:
Added tests for creation of Attachments.

jon-wobken:
added base64 support #485 v1.11.5

PR #298, opened 2019-04-12
Update tests for apexecute

courtneyod:
Added test coverage for `apexecute`


cgabard:
With a recent update, our SF instance start complaining and we find the error narrow to this point.

We have implemented a workaround but I would prefer have a clean solution. I see that there is other attempt to fix this (see https://github.com/simple-salesforce/simple-salesforce/pull/324 for example).

Can I provide help to fix this and have it merged and released ? It seems the build issue is mainly a line too long.

PR #303, opened 2019-05-17
added timeout option

warrenisarobot:
The default timeout option for requests is None, so that if salesforce is down the request will continue indefinitely.  This lets us specify a default timeout for the request, like:

sf = simple_salesforce.Salesforce('username', 'password', 'token', timeout=30)

So that all requests for this `sf` object will use the default timeout of 30.  This only uses the default timeout on Salesforce._call_salesforce() if `timeout` was note passed as a parameter, and if a default timeout was specified


pdr:
Very very timely!!! Love it.

xmedeko:
As a workaround, you may hack the requests session:
```py
session = requests.Session()
session.request = functools.partial(session.request, timeout=60 * 30)
sf = Salesforce(..., session=session)
```

PR #305, opened 2019-05-24
Add support for JWT Bearer Token Flow

wgordon17:


wgordon17:
@Mer-xes or @nickcatal can either of you review and merge?

PR #314, opened 2019-08-27
Handling IndexError on empty response

finomayato:
If `result.json()` is empty: `result.json()[0]` will raise `IndexError`


finomayato:
Actually, the root cause will be that we don't properly handle Failure on query bulk requests. Will try to implement a fix for it a little bit later

Issue #351, opened 2020-02-02
cloudforce.com login URL support


madeleineth:
I am trying to use simple-salesforce with a domain whose login URL ends with `cloudforce.com`, not `salesforce.com`. I've confirmed that it works fine if I patch [this line in login.py](https://github.com/simple-salesforce/simple-salesforce/blob/master/simple_salesforce/login.py#L76). I'd be happy to send a PR to add a general solution, but I'd like to get a sense of what the maintainers would like, design-wise.

Some options:
1. Change the semantics of `domain` to somehow accept fully-qualified domain names in addition to the partial ones it takes now. Somehow you'd need to tell them apart. I'm not sure what the rule would be. To make things interesting, `.my` is the Malaysia TLD. So `u.com.my` is a fully-qualified name but `ucom.my` could be a Salesforce corporate domain. Or we could special-case `cloudforce.com`. I don't know whether there are ever login domains that don't end in either `salesforce.com` or `cloudforce.com`. This one makes me nervous because I don't want to ever guess wrong when deciding what hostname to send login credentials to.
2. Deprecate the `domain` argument and add a `login_domain` argument to the `Salesforce` and `SalesforceLogin` objects. Specifying `domain=x` would be equivalent to specifying `login_domain=(x + '.salesforce.com')`. 
3. Same as (2) except don't deprecate `domain`. Indefinitely allow at most one of `sandbox`, `domain`, and `login_domain` to be specified.

Would y'all take a PR to do any of these, or something else?

jon-wobken:
I can see some value in this. I think it would be better to change how the {domain} argument works...I like the idea of it accepting to full url (ex. domain.salesforce.com // domain.cloudforce.com // corporate_name.my.salesforce.com).  Though, this would require updating in several places.

I believe after logging in the instance_url is set from the response so there shouldn't be an issue once the login url has been changed. The domain argument is used in several other places so I'd be hesitant on changing how it works without taking a further dive into it. I think 3 would be the best option in my mind. There is already conditions to bypass the sandbox argument if domain and test are both set so I'd recommend something similar and setting the full instance url if the login_domain is not none, else fall back to domain, else fall back to default.

madeleineth:
If anyone cares to see where I'm going with #353, look at my [c1 branch](https://github.com/madeleineth/simple-salesforce/tree/c1). I'm splitting into two PRs to make review easier.

Kevmo92:
👍 on the idea to accept a full_domain

[...]

PR #353, opened 2020-02-08
Use _test_login_success to simplify test_login.py.

madeleineth:
There was a lot of repeated code in `test_login.py`, and I was going to have to add even more in a test of #351. So, this commit pulls out the common code for succesful logins into `_test_login_success`, which is called by other tests. `test_custom_session_success` goes away entirely because `test_default_domain_success` is almost identical.


madeleineth:
This is my first time as a contributor to `simple-salesforce`, so for the record, neither I nor my employer claim any copyright on the content of this PR.

jon-wobken (approved):

Issue #257, opened 2018-06-13
Lazy loading in `query_all`


kokes:
Hi, I have been working with larger data exports on memory-constrained systems, so I wanted to make simple-salesforce lazy. I have a local wrapper around `query` to mimick `query_all`, but in a way it iterates over the returned records across all pages. There are a few upsides:

1. We don't have to wait for all pages to get loaded, so it has lower latency
2. We don't load all records at once, so the memory pressure is lower
3. If we stop our iterator halfway through, we save ~half our API calls, because they won't get fired.

There is at least one downside: you don't know the total number of records until the end of the iteration. We know the number (from the API call), but it would be very clunky to include it in the iterator (it would have to be coupled with every yielded record). The only other option is to return the number of records and an iterator (in a tuple) and then one could use either. Something along the lines of

```
n, data = sf.query_all_iter(...)
logging.info('looping %d records', n)
for el in data:
  process(el)
```

Anywho, the original design, where the method just iterates over the values returned is shown here, together with a diff over the original `query_all` method, just so you see that it's almost a 1:1 copy.

https://gist.github.com/kokes/706e35951febe0994219faef0b00d36d/revisions#diff-1ef5aa43d162ea30e45b40e5bcdd0748

andscoop:
@kokes Thanks for the example, would you be interested in opening a PR for this generator functionality? Ideally moving combined the logic of .query_all and .query_all_iter into a private method, which query_all and query_all_iter will interface with

luisvillarreal:
@andscoop @kokes Is this `query_all_iter` available now. I'm dealing with a large volume of data and would like to stream all the results to an on-prem HDFS.

luisvillarreal:
@kokes This is real good!

ddow:
This is they greatest thing to happen to SS since the bulk API integration.

jon-wobken:
@luisvillarreal - I have not seen a pull request for this. 

@kokes - would you be interested in opening a PR for this?

kokes:
I will look into it later this week. I'll think a bit about the API - maybe the iteration functionality doesn't have to be private - it can be a public API that will be leveraged by `query_all` directly. Should be simple enough.

luisvillarreal:
This works really good. I just downloaded 1 GB of data to HDFS. I also had memory constrains.

Kudos @kokes !

kokes:
[...]

PR #354, opened 2020-02-12
Lazy loading

kokes:
To ensure backward compatibility, we're adding a new method (`query_all_iter`), which is then reused in a new implementation of `query_all`.

Open questions:
- do we need to duplicate docs in README.rst and queries.rst? I just copied it as I saw it in the existing docs.
- ~~what does `query_all` return against a live Salesforce instance? I'll test it tomorrow once I get the chance - I'm not sure if totalSize is a part of the return object or not (the tests don't suggests so, but the docs do)~~

For some context, see #257

### DIFF
--- a/simple_salesforce/tests/test_api.py
+++ b/simple_salesforce/tests/test_api.py
@@ -724,7 +724,8 @@ class TestSalesforce(unittest.TestCase):
             responses.GET,
             re.compile(r'^https://.*/queryAll/\?q=SELECT\+ID\+FROM\+Account$'),
             body='{"records": [{"ID": "1"}], "done": false, "nextRecordsUrl": '
-                 '"https://example.com/queryAll/next-records-id", "totalSize": 2}',
+                 '"https://example.com/queryAll/next-records-id",'
+                 '"totalSize": 2}',
             status=http.OK)
         responses.add(
             responses.GET,
--- a/simple_salesforce/tests/test_api.py
+++ b/simple_salesforce/tests/test_api.py
@@ -660,6 +660,33 @@ class TestSalesforce(unittest.TestCase):
         self.assertEqual(result, {})
 
     @responses.activate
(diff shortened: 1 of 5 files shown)


jon-wobken:
@kokes - Overall looks good, just going to test in a few of my live instances and then I'll merge. Definitely a great addition - looking forward to using it.

simple_salesforce/tests/test_api.py
```diff
@@ -705,7 +732,7 @@ def test_query_all(self):
             OrderedDict([(u'records', [
                 OrderedDict([(u'ID', u'1')]),
                 OrderedDict([(u'ID', u'2')])
-            ]), (u'done', True)]))
+            ]), (u'done', True), (u'totalSize', 2)]))
```
jon-wobken, line 735: Can remove "u" from OrderedDicts to be consistent with current test_api convention and removing support for python 2.*

simple_salesforce/tests/test_api.py
```diff
@@ -735,7 +763,7 @@ def test_query_all_include_deleted(self):
...
```
jon-wobken, line 766: Can remove "u" from OrderedDicts to be consistent with current test_api convention and removing support for python 2.*

simple_salesforce/tests/test_api.py
kokes: rebased on the current master and resolved this

simple_salesforce/tests/test_api.py
kokes: rebased on the current master and resolved this

kokes:
Ended up rebasing on the current master, so it's all current - the only conflicts were in tests (with all the `u'` stuff, so I resolved that and that resolved the points you raised as well.

jon-wobken:
@kokes - Merged.

PR #374, opened 2020-04-03
Send an empty body, not "null" if no data is specified.

edanaher:
This tweaks the apexecute method to handle and empty body/data correctly; previously, if the `data` passed into `apexecute` was None, it would be passed through json.dump, production `"null"`.  This data was then passed through the requests library, which happily added a "Content-Length: 4" header which caused Salesforce to (at least in some cases) reject the request.

The solution is simply to check for None, and pass in None explicitly rather than json-ifying it.

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -434,11 +434,14 @@ class Salesforce:
         * data -- A dict of parameters to send in a POST / PUT request
         * kwargs -- Additional kwargs to pass to `requests.request`
         """
+        # If data is None, we should send an empty body, not "null", which is
+        # None in json.
+        json_data = json.dumps(data) if data is not None else None
         result = self._call_salesforce(
             method,
             self.apex_url + action,
             name="apexexcute",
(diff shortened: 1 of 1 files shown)


edanaher:
Some additional background:

We're using simple-salesforce to sync contacts to Salesforce via a mapping defined in Salesforce.  To load the mapping into our application, we're using a custom application endpoint, conveniently accessible via apexecute here.

However, in at least some cases, we're getting a mysterious HTTP 411 indicating Salesforce is under maintenance.  Experimentally, this seems to be due to sending a "Content-Length: 4" header in the request to Salesforce.  This is a bug, since a GET request shouldn't have any content or Content-Length.

Since it's such a simple fix, I figured it's less effort to open a PR with my fix than to submit an issue describing the problem.

cesar-rivera-adroll:
Hey guys, sorry to bother, this issue is also affecting us and this fix would be very useful, any timeline as to when this will be merged and a new version be available?

PR #384, opened 2020-05-23
Fix supported Python version in docs.

jtroup:
Update the docs to match reality; the 1.0 release dropped support for Python 2.6, 2. 7, 3.3 and 3.4.


jtroup:
I'm not sure where to report this but, as an aside, the package on pypi also claims to support Python 2 but doesn't.  Specifically https://pypi.org/project/simple-salesforce/#files points at a simple_salesforce-1.0.0-py2.py3-none-any.whl which doesn't actually work with Python 2.

Issue #389, opened 2020-06-01
Enhancement: Quoting/escaping for SOQL values and external IDs


countergram:
To effectively use `query()` at all or use external IDs with `upsert()` or `get()`, quoting/escaping must be applied. Think it would be good for this to be built into the library, either as helpers or built into these functions. Otherwise it has to be redone in every implementation.

Willing to contribute a PR. Helpers would be easy... looks like building into `query()` would be possible though it already uses `**kwargs` for something else so it would need to be a specific named argument.

PR #394, opened 2020-06-07
Add format_soql and format_external_id functions

countergram:
Resolves https://github.com/simple-salesforce/simple-salesforce/issues/389. This adds a couple of simple utility functions to perform needed quoting & escaping in order to prevent errors or injections. Tests & docs included.

While this started out with string quoting, which I implemented in a work project and have been running, I quickly realized the need to include other types, as well as the need for 2 additional cases: `literal` exists to prevent users from having to use nested format strings to add things like dynamic object names in other ways, and `like` exists because, given an entire right side value for a like expression, it's not possible to tell which `_` and `%` characters are intended to be wildcards and which are part of a dynamic input.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -181,6 +181,26 @@ While ``query_all`` materializes the whole result into a Python list, ``query_al
     for row in data:
       process(row)
 
+Values used in SOQL queries can be quoted and escaped using ``format_soql``:
+
+.. code-block:: python
+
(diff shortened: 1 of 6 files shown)


cwlms:
Really useful honestly.  Well done!

jon-wobken:
@countergram - Looks great and very helpful.

PR #409, opened 2020-07-29
Add parameter to allow JWT expiry in local timezone

cth1011:
This allows the users to utilize the JWT method in local timezone instead of UTC

### DIFF
--- a/simple_salesforce/login.py
+++ b/simple_salesforce/login.py
@@ -8,7 +8,7 @@ DEFAULT_CLIENT_ID_PREFIX = 'RestForce'
 
 import time
 import warnings
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from html import escape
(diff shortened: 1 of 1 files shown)


simple_salesforce/login.py
```diff
@@ -154,7 +155,10 @@ def SalesforceLogin(
             consumer_key is not None and \
             privatekey_file is not None:
         header = {'alg': 'RS256'}
-        expiration = datetime.utcnow() + timedelta(minutes=3)
+        dt = datetime.utcnow()
+        if set_exp_to_local:
```
jon-wobken, line 159: Could consolidate these three lines to a singular line

`dt = datetime.now(timezone.utc).astimezone() if set_exp_to_local else datetime.utcnow()`

Other than that I am good to merge this

jon-wobken:
Duplicated by #431

PR #413, opened 2020-08-19
Explicitly encode XML as UTF-8.

by-mb:
I had the problem my password contained multi-byte UTF-8 characters. Unfortunately if you pass a string as the request body to a requests POST, it seems to (wrongly) encode this as Latin-1 encoding (see also: https://github.com/psf/requests/issues/4503). As this is unexpected and XML declaration states it should be UTF-8, it should also be explicitly encoded as UTF-8.

### DIFF
--- a/simple_salesforce/login.py
+++ b/simple_salesforce/login.py
@@ -103,7 +103,7 @@ def SalesforceLogin(
             </env:Body>
         </env:Envelope>""".format(
             username=username, password=password, token=security_token,
-            client_id=client_id)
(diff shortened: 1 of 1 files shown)


by-mb:
@jon-wobken any chance getting this bugfix merged?

by-mb:
@jon-wobken More than a year now. I think the fix is still valid.

by-mb:
@jon-wobken I once again brought this one up to date with the master branch. Now would be a perfect time to merge it. ;-)

by-mb:
I think more than 5 years is enough waiting time. Closing this one.

PR #427, opened 2020-10-28
Issue #375: Feature Request: File Based Metadata API Calls

cmitzel:
Documentation for use provided in README.RST

Following up on @Kevmo92's feature request for File Based Metadata API Calls (https://github.com/simple-salesforce/simple-salesforce/issues/375)

- Ported source code from https://github.com/Kevmo92/sfdclib
   - `messages.py`: XML message templates for Salesforce API
   - `metadata.py`: functionality to deploy zipfile and check deployment status
   - `session.py` -(renamed)> `sfdc_session.py`: existing session class to deal with mdapi interaction

- Added functionality to Salesforce class 
   - examples for using deploy() and checkDeployStatus() available in README.RST

- Added unit tests to test_api.py
   - Covers various cases for deploy() and checkDeployStatus()

- Additional testing was done to ensure compatibility with @nCino's internal tools.

Issue #375, opened 2020-04-09
Feature Request: File Based Metadata API Calls


Kevmo92:
It would be great if simple-salesforce supported making [file based calls](https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_file_based_calls_intro.htm).

Ideally we could make a call like:

```python
from simple_salesforce import Salesforce

deployment_finished = False
successful = False

sf = Salesforce(session_id="id", instance="instance")
aysncId, state = sf.deploy("path/to/zip", options)

while not deployment_finished:
    status = sf.checkDeployStatus(asyncId)
    if status.get("state") in ["Succeeded", "Completed", "Error", "Failed", None]:
        deployment_finished = True
    if status.get("state") in ["Succeeded", "Completed"]:
        successful = True

if successful:
    print("✅")
else:
    print("🥔")
```

There's an older project on GitHub that supports these file based calls [here](https://github.com/rbauction/sfdclib), but it would cool if simple-salesforce supported them as well.

Selfishly, I'm more concerned about the deploy and checkDeployStatus methods.

PR #434, opened 2020-12-04
Issue #375: Feature Request: File Based Metadata API Calls V2

jon-wobken:
Re-factor from #427 

Thanks @cmitzel for contributions


cmitzel:
Hi @jon-wobken, nCino's been using the release branch of simple_salesforce and found recently that the `sandbox` variable in the md deploy breaks a few things, so we've reverted to my forked branch in the meantime. What was your reasoning behind requiring it, or not having a default value, for md deploy here?

jon-wobken:
@cmitzel  - I'll check that out. The main reason for requiring this is do to the allowMissingFiles and rollbackOnError argument that can be passed through kwargs. Salesforce notes not to set allowMissingFiles argument for deployment to production orgs and to set rollbackOnError to true for production orgs since its required https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_deploy.htm. I handle this in the metadata package and set these using:

`if not self._sandbox:
            attributes['allowMissingFiles'] = False
            attributes['rollbackOnError'] = True`

Though - as I re-read the documentation it might be wiser to delete that from the xml deploy message if that is the case.

PR #435, opened 2020-12-13
full_domain argument to constructors

madeleineth:
This lets you log in on domains other than ones ending in ".salesforce.com".

Also drop a redundant test, `test_custom_session_success`, which runs the same code as `test_default_domain_success`.

Close https://github.com/simple-salesforce/simple-salesforce/issues/351.

I have tested this locally with `tox` and against a Salesforce sandbox with a non-salesforce.com domain.


brno32:
This looks neat!

madeleineth:
This PR has been sitting here for a while, so I rebased onto `master`, re-ran the tests, and fixed a resource-deallocation issue elsewhere in the package. I've been running my fork as part of a deployed application for a few months now.

Issue #442, opened 2021-01-10
Bulk API silently fails when a batch fails due to invalid json


tsabat:
I use pandas and simple salesforce.  I was pushing bulk objects to SFDC from a Pandas dataframe that occasionally had a `pd.nan` in it.  The problem here is twofold:

### Problem 1: `NaN` is sent to SFDC during `json.dump`

1. `json.dumps` has an `allow_nan` property which [defaults to true](https://stackoverflow.com/questions/6601812/sending-nan-in-json/6602204#6602204)
2. `NaN` is not part of the json spec
3. SFDC chokes when it receives a `NaN`

### Problem 2: Batches with invalid json silently fail.

If you have > 200 values in your bulk call, it's split into groups of 200 by the simple salesforce lib and sent in batches.  If your first group successfully updates (because it has no `NaN` in it) but your second group fails (because it does have `NaN` in it) the errors result is truncated to the first group.  An example is below.

this code 

```python
account_length = len(data["ship_to"])
print(f"pushing {account_lenth} ship_to accounts")
results = client.bulk.Account.upsert(data["ship_to"], "ACID__c")
print(len(results))
```

results in this output

```bash
pushing 1150 ship_to accounts
200
```

That's because the batch has invalid json.  You can see it in the bulk UI on the salesforce side.

![nan](https://user-images.githubusercontent.com/636041/104222235-de34e200-53f6-11eb-97e9-b5351a536979.png)

### Suggested fix

When you [dump the payload](https://github.com/simple-salesforce/simple-salesforce/blob/master/simple_salesforce/bulk.py#L103) you should use the set the [allow_nan](https://docs.python.org/3/library/json.html) to `False` like so:

```python
        result = call_salesforce(url=url, method='POST', session=self.session,
                                 headers=self.headers,
                                 data=json.dumps(payload, allow_nan=False))
```

PR #443, opened 2021-01-11
prevents silent failure of bulk API by preventing NaN in payload.

tsabat:
calling `json.dumps` with `allow_nan=True` (which is the default) results in a silently failing batch.  This patch fixes https://github.com/simple-salesforce/simple-salesforce/issues/442 by calling `json.dumps` with `allow_nan=False`.  This will cause code with `NaN` in it to fail fast, rather than silently if involved in a large-sized batch.

### DIFF
--- a/simple_salesforce/bulk.py
+++ b/simple_salesforce/bulk.py
@@ -102,7 +102,13 @@ class SFBulkType:
 
         result = call_salesforce(url=url, method='POST', session=self.session,
                                  headers=self.headers,
-                                 data=json.dumps(payload))
+                                 data=json.
+                                 
+                                 
+                                 
+                                 
+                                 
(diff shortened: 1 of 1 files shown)


jon-wobken:
Great find. Never experienced this error. I'll look to review and merge this week.

Phlogi:
SFDC chokes because the json is not correct, it should look like this in the request: 

```
{
        "field": null,
}
```

Now we need to make sure that json.dumps produces this null values correctly, which this change here helps i guess.

I'm using pandas to work with simple-salesforce and as long as I ensure python None-Type is used for null's it works as expected:

` data.replace({np.nan: None}).to_dict('records')` 

This dict is then passed to the bulk insert function and works as expected.

PR #451, opened 2021-02-19
Support signing JWT without requiring a file to read the key from

andrewphahn:
Adds the ability to pass a private key as `bytes` or `str` when using the JWT auth flow. This makes it so the key does not need to reside on a disk accessible to the code, which is especially useful for serverless deployments.

- `privatekey` parameter added to `Salesforce` and `SalesforceLogin`

`privatekey_file` takes precedence over `privatekey` if both are passed.

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -49,6 +49,7 @@ class Salesforce:
             domain=None,
             consumer_key=None,
             privatekey_file=None,
+            privatekey=None,
             ):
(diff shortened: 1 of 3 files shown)


andrewphahn:
@jon-wobken any chance you could take a look at this one and let me know if it is on the right track?

jon-wobken:
@andrewphahn - Looks good from my side. Like that the file takes precedence as a precaution if both are passed.

Issue #145, opened 2017-01-12
set_password returns ValueError exception when successful


nicholas-brown:
When set_password() completes successfully, it returns a ValueError exception with value 'No JSON object could be decoded'

skamensky:
This is a result of the response being empty. I've tested this and also according to Salesforce documentation:

> No response body on successful password change, HTTP status code 204 returned.

According to requests documentation

> In case the JSON decoding fails, r.json() raises an exception. For example, if the response gets a 204 (No Content), or if the response contains invalid JSON, attempting r.json() raises ValueError: No JSON object could be decoded.

I'm not sure if another error is occurring because we've set the conversion from dicts to ordered ditcs here https://github.com/simple-salesforce/simple-salesforce/commit/99559e009117c9c3cc47c2882297246bbf529a7f.

I think we just need to catch the ValueError and treat it as a success.

tstarbow-sfdc:
I am seeing this issue too. Call works correctly, but ValueError is raised.

espoelstra:
I'm also seeing this. I was breaking my brain trying to figure out where it was coming from when I was writing unit tests around my simple_salesforce CLI wrapper and I'd forgotten that the code was rather goofy in that spot. It seems silly trying to load the json_results when there aren't any expected per the API.

espoelstra:
Even more fun, the error messages are different based on whether you are using Python 2.7.x or 3.7.x. :-P
```
    # Python 2
    assert 'No JSON object could be decoded' in result.output
    # Python 3
    assert 'Expecting value: line 1 column 1 (char 0)' in result.output
```

PR #473, opened 2021-04-26
Fix set password to handel empty response on success

bastir85:
Fixes https://github.com/simple-salesforce/simple-salesforce/issues/145

The REST API of salesforce returns a 204 status code without any data on successful password change. This is a minimal change from the previous implementation. However, it looks like the json parsing for the status code 200 can also be removed. 

https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_sobject_user_password.htm


bastir85:
@jon-wobken Does this look good to you, would be glad if it can be merged.

PR #485, opened 2021-05-27
Added upload_base64, update_base64, get_base64 functionality

adam-mah:


mhogg-srg:
This is great functionality, but I think the naming convention needs changing. I'm thinking that insert_file, update_file, and get_file might be better choices. My reasoning:

1) From the calling code doesn't need to know anything about base64.

2) It's my understanding that when inserting/updating data into the Attachment object (for instance), the bytestring needs to be base64 encoded. However, when retrieving the body data with a call to a path like sobjects/Attachment/ABC123/Body, the binary data is NOT base64 encoded. Therefore get_base64 isn't returning any base64 data.

Also it might be preferable to create more abstract methods that take bytestring arguments for the CRU operations, then call those methods from the methods that read/write disk files.

PR #507, opened 2021-09-17
FEAT: refresh sessions automatically

NoahCardoza:
This is a start to addressing the issue outlined in #292.

While I think it should work for every type of login, I'm not sure it's the _best_ way. I'm pretty new to Salesforce and I'm pretty sure I've seen some documentation about actual refresh tokens which I have not incorporated.

This PR also needs tests set up. While I started, I don't have too much experience writing good tests and don't have much time at the moment. However, I can get to that later. I figure it would be more helpful to get this out and figure out testing in the future.

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -7,6 +7,7 @@ import json
 import logging
 import re
 from collections import OrderedDict, namedtuple
+from functools import partial
 from urllib.parse import urljoin, urlparse
 
 import requests
@@ -14,8 +15,8 @@ import requests
 from .bulk import SFBulkHandler
 from .exceptions import SalesforceGeneralError
 from .login import SalesforceLogin
-from .util import date_to_iso8601, exception_handler
 from .metadata import SfdcMetadataApi
+from .util import date_to_iso8601, exception_handler
 
 # pylint: disable=invalid-name
 logger = logging.getLogger(__name__)
@@ -107,6 +108,7 @@ class Salesforce:
         self.domain = domain
         self.session = session or requests.Session()
         self.proxies = self.session.proxies
+        self._salesforce_login_partial = None
         # override custom session proxies dance
         if proxies is not None:
             if not session:
@@ -124,7 +126,8 @@ class Salesforce:
             self.auth_type = "password"
 
             # Pass along the username/password to our login helper
-            self.session_id, self.sf_instance = SalesforceLogin(
+            self._salesforce_login_partial = partial(
+                SalesforceLogin,
                 session=self.session,
                 username=username,
                 password=password,
@@ -133,6 +136,7 @@ class Salesforce:
                 proxies=self.proxies,
                 client_id=client_id,
                 domain=self.domain)
+            self._refresh_session()
 
         elif all(arg is not None for arg in (
                 session_id, instance or instance_url)):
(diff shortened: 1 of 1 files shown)


simple_salesforce/api.py
```diff
@@ -207,6 +212,20 @@ def __init__(
 
         self.api_usage = {}
 
+    def _generate_headers(self):
```
NoahCardoza, line 215: Extract this to a new method since it will be used in two places.

simple_salesforce/api.py
```diff
@@ -207,6 +212,20 @@ def __init__(
...
+        self.headers = {
+            'Content-Type': 'application/json',
+            'Authorization': 'Bearer ' + self.session_id,
+            'X-PrettyPrint': '1'
+        }
+
+    def _refresh_session(self):
```
NoahCardoza, line 222: Run the login logic cached in `__init__` and recalculate the headers with the new `session_id`.

simple_salesforce/api.py
```diff
@@ -124,7 +126,8 @@ def __init__(
             self.auth_type = "password"
 
             # Pass along the username/password to our login helper
-            self.session_id, self.sf_instance = SalesforceLogin(
+            self._salesforce_login_partial = partial(
```
NoahCardoza, line 129: I'm basically using `partial` to cache the values passed to `SalesforceLogin` in `_salesforce_login_partial` so later when our session expires, we can just call `SalesforceLogin` again and get a new token.

simple_salesforce/api.py
```diff
@@ -148,13 +152,17 @@ def __init__(
                     self.sf_instance += ':' + str(port)
             else:
                 self.sf_instance = instance
+            
+            # Only generate the headers wihtout logging in first
+            self._generate_headers()
```
NoahCardoza, line 157: Since we don't actually perform a login here, we have to remember to generate the headers (usually done in the `_refresh_session` method).

simple_salesforce/api.py
```diff
@@ -679,6 +713,12 @@ def __init__(
                                      object_name=object_name,
                                      sf_version=sf_version))
 
+    @property
+    def session_id(self):
```
NoahCardoza, line 717: I use a property here to make it backward compatible with the tests already in place. However, if this officially gets merged in we'd probably want to just pass most everything through the `salesforce` parameter.

shanehh:
Hi simple-salesforce team.. please merge this PR?

dansmachina:
Hi! Following up on this feature, it would be really nice to have it 🥇

simple_salesforce/api.py
```diff
@@ -129,7 +126,8 @@ def __init__(
             self.auth_type = "password"
 
             # Pass along the username/password to our login helper
-            self.session_id, self.sf_instance = SalesforceLogin(
+            self._salesforce_login_partial = partial(
```
jon-wobken, line 129: Should be a quick addition but should ensure that it supports the OAuth and Security token login options

- [ ] Pass through OAuth arguments
- [ ] Pass through security token arguements

NoahCardoza: Just saw this! Summer is coming up soon and I'll likely resume the project I was using this for and thus have time to take a look at this. I'll look into it and come back if I have any questions.

Thanks for taking a look!

PR #520, opened 2021-12-21
Add auto-completion for SObject names

mhogg-srg:
Salesforce objects retrieve SObject names during initialization and append them to the normal elements returned by __dir__, allowing for auto-complete of SObject names found in the Salesforce instance. (Sadly, not all IDEs support auto-completion using __dir__.)

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -207,6 +207,12 @@ class Salesforce:
 
         self.api_usage = {}
 
+        try:
+            self._sobject_names = [x['name'] for x
+                                   in self.describe()['sobjects']]
+        except: # pylint: disable=bare-except
+            self._sobject_names = []
+
     def describe(self, **kwargs):
         """Describes all available objects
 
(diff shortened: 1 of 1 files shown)


paneidos:
This would cost one API call per instance of the Salesforce class.

mhogg-srg:
> This would cost one API call per instance of the Salesforce class.

That's an excellent point. I guess that makes this idea undesirable, unless an optional parameter were added to the constructor, something like a boolean called `interactive`.

simple_salesforce/api.py
```diff
@@ -207,6 +207,12 @@ def __init__(
 
         self.api_usage = {}
 
+        try:
+            self._sobject_names = [x['name'] for x
+                                   in self.describe()['sobjects']]
+        except: # pylint: disable=bare-except
+            self._sobject_names = []
+
```
ericbn, line 215: Maybe this code could be moved inside the `__dir__` function, so it does not run in production code. Also not sure the [bare except](https://stackoverflow.com/a/4992124/2654518) is a good idea -- could be at least `except Exception:`.

PR #521, opened 2021-12-21
Added Tableau CRM(Einstein Analytics/Wave Analytics) API support

amd-yan:
Not sure if this is a common use case. We load SFDC data to Snowflake on an hourly basis, prep the data for Tableau CRM dashboard in Snowflake, then Sync back to Tableau CRM. 

However, I can't find any ways to orchestrate this data pipeline. Therefore, I added Tableau CRM APIs to this cool package, so that we can orchestrate everything in Apache Airflow. 

This is the initial implementation of the Tableau CRM APIs. I only included the get and post methods. There are other functions, such as patching the records, that are not implemented (yet).

PS: I also bumped up the default API version to 53 from 42, because many APIs require 49 or above. Hopefully, it is not an issue.

Let me know should you have any other questions. Thanks for maintaining this awesome tool!


amd-yan:
PS: the build failure is not from code because it shows the same error on the master branch:
[https://app.travis-ci.com/github/simple-salesforce/simple-salesforce/builds](https://app.travis-ci.com/github/simple-salesforce/simple-salesforce/builds)
After some quick google search, it seems that cryptography >=3.5 starts to require rust: [https://github.com/pyca/cryptography/issues/5771#issuecomment-775016788](https://github.com/pyca/cryptography/issues/5771#issuecomment-775016788)

jon-wobken:
I'll need to dig into this a bit more. I don't have an environment to validate/test functionality as I do not have access to a Tableau CRM enabled org so I will need to investigate and look to include this on my next set of items I review.

amd-yan:
Thanks for your heads up!  Please let me know if you have any questions/concerns. Since I didn't find any better ways to orchestrate the jobs between Snowflake and Tableau CRM, I have been using this combined with Airflow to do so. There are many other APIs for Tableau CRM that I have not implemented yet. If there are any good use cases, I am happy to spend some time and add those too.

amd-yan:
Hi @jon-wobken  - I hope everything is going well. A few weeks ago, I spent some time to merge this Wave API to your latest release (1.12.6). Salesforce probably have added new functions to the Wave API since I wrote this in 2021, we have been using it to schedule/trigger the CRMA (TCRM) jobs for the last 4 years. If I have time, I will revisit this to check if we can add more functions to it later. Let me know if you have any questions/concerns.

Issue #506, opened 2021-09-17
SOAP API 7 is retired, but likely still in use even 1.11.4


RedHair-CA:
Dear friends,
Our SF admin informed me to make a code change ASAP due to the retirement of SOAP 7.0.
And their API log indicated such a log is from my program which leads to simple_salesforce as per my analysis. 
It looks like in API.py, line 203, a SOAP call is used.

My code to call with that API log, I need the sf.session_id for next codes.

sf = Salesforce(username=sfusername,
                          password=sf_pwd,
                          security_token=sf_token,
                          domain=domain,
                          version="v51.0")

Could you please help to update the SOAP version please? 

Thank you a lot!

Ref link:
Salesforce Platform API Versions 7.0 through 20.0 Retirement  --  https://help.salesforce.com/s/articleView?id=000351312&type=1
Log sample: 
![image](https://user-images.githubusercontent.com/46230780/133816018-4a554b18-70dc-4215-ae8d-83189932cde0.png)

RedHair-CA:
It has been likely 2 months, may anyone comment a bit please? This sounds likely a wildly impacted thing, isn't it?

hkajiwara:
@RedHair-CA 

Sounds weird to me. The latest version of this library leverages '42.0' as [DEFAULT_API_VERSION](https://github.com/simple-salesforce/simple-salesforce/blob/30ad772f81935b604b6461e17e1c92c16f77f9b2/simple_salesforce/api.py#L4). Actually, Login History record shows '42.0' when I log into a test org. Probably, you would use a quite old library or specify the old version in your code.

<img width="381" alt="Screen Shot 2021-11-24 at 23 11 34" src="https://user-images.githubusercontent.com/4019992/143255390-c443ca2e-506f-4d9e-9b72-24998a055d81.png">

RedHair-CA:
Thank you so much for reply hkajiwara!
I use simple-salesforce==1.11.4
And assign Salesforce api version 51.0 actually when create connection. But the issue is in deeper part of the source code of simplesalesforce, where I found via debug.
I can see in line 203 of api.py - self.metadata_url = ('https://{instance}/services/Soap/m/{version}/'
                             .format(instance=self.sf_instance,
                                     version=self.sf_version))
the self.sf_version = 'v51.0'
That's the only place I saw SOAP is called, so doubt that line.
My other SF program without SimpleSalesforce doesn't have issue on SOAP v7 at all... 
[...]

PR #522, opened 2021-12-22
Strip leading v from api_version

hamiltonkibbe:
This is related to #506, and  is really more of a fix for a salesforce bug than a library bug.

The behavior that I've noticed is that specifying a version in the SOAP url  with a leading 'v', such as 'v51.0', salesforce returns a server_url with version '7.0' which is soon to be retired. I'd expect the specifying an invalid version in the login url to result in an error from salesforce rather than it being silently thrown out and replaced with a soon-to-be-retired version, which is really surprising behavior from a user perspective. 

This PR's approach is one option for working around that behavior. This is one approach that doesn't change the library interface at all, If you'd prefer this to just raise a ValueError if the user specifies an invalid api_version -- which is arguably more correct -- I can update this PR to use that approach as well.

### DIFF
--- a/simple_salesforce/login.py
+++ b/simple_salesforce/login.py
@@ -65,8 +65,10 @@ def SalesforceLogin(
 
     if domain is None:
         domain = 'login'
-
-    soap_url = 'https://{domain}.salesforce.com/services/Soap/u/{sf_version.lstrip("v")}'
+    
+    sf_version = sf_version.lstrip("v")
+    
+    soap_url = 'https://{domain}.salesforce.com/services/Soap/u/{sf_version}'
 
     if client_id:
         client_id = "{prefix}/{app_name}".format(
--- a/simple_salesforce/login.py
+++ b/simple_salesforce/login.py
@@ -66,7 +66,7 @@ def SalesforceLogin(
     if domain is None:
         domain = 'login'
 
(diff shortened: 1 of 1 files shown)


jon-wobken:
> This is related to #506, and is really more of a fix for a salesforce bug than a library bug.
> 
> The behavior that I've noticed is that specifying a version in the SOAP url with a leading 'v', such as 'v51.0', salesforce returns a server_url with version '7.0' which is soon to be retired. I'd expect the specifying an invalid version in the login url to result in an error from salesforce rather than it being silently thrown out and replaced with a soon-to-be-retired version, which is really surprising behavior from a user perspective.
> 
> This PR's approach is one option for working around that behavior. This is one approach that doesn't change the library interface at all, If you'd prefer this to just raise a ValueError if the user specifies an invalid api_version -- which is arguably more correct -- I can update this PR to use that approach as well.

I'd rather have a value error raised just to provide context back to the user

cscetbon:
Anyone can check why Travis is failing and how to get it fixed and merged in ?

hamiltonkibbe:
As far as I can tell, it's failing due to missing dependencies on the build agent: cryptography install depends on a rust compiler being available as of early 2021-ish and there is no rust compiler available when the tox tests run on python 3.8

jon-wobken:

PR #530, opened 2022-02-09
Added the ability to stream results from the bulk API.  

bkbeckman:
I have been using simple-salesforce to pull information from our very large SFDC instance.  I discovered that the bulk API implementation can allocate large amounts of memory when pulling extremely large result sets.

This commit provides the ability to stream results from the bulk api, which keeps the memory footprint low.  

I have provided two mechanisms for streaming from the bulk API, one is streamed JSON, which uses the json-stream library, and the other is streamed plain text.

### DIFF
--- a/docs/user_guide/queries.rst
+++ b/docs/user_guide/queries.rst
@@ -67,3 +67,23 @@ Search and Quick Search return ``None`` if there are no records, otherwise they 
 More details about syntax is available on the `Salesforce Query Language Documentation Developer Website`_
 
 .. _Salesforce Query Language Documentation Developer Website: http://www.salesforce.com/us/developer/docs/soql_sosl/index.htm
+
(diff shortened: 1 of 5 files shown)


jon-wobken:
Does the existing lazy_operation that returns a generator handle the memory issue you are facing?

bkbeckman:
No, because lazy_operation still must download the entire contents of the request and parses it before it can begin to return results.

This PR does two things:
1. Passes the stream=True flag to the session.request method, which prevents the entire request content from being downloaded before being returned.
2. Adds the json-stream library, which is capable of parsing json without loading the entire string into memory first.

jon-wobken:

bkbeckman:
I appreciate it.  Let me know if there's anything I can assist with.

PR #545, opened 2022-04-11
Implement wait param in non-query bulk operations

Haeilifax:
Add the wait parameter to workers spawned from _bulk_operation for non-query calls (it already exists for query calls). This should have no impact in 99% of cases, as the wait parameter is not exposed on higher-level calls, but allows for manual control over response times for the highly-motivated.


jon-wobken:
I’ll review this shortly - you can ignore the Python 3.5 Travis-CI. I’ll likely drop support as it is almost two years deprecated and add support for 3.9/3.10

jon-wobken:

jon-wobken (approved):

Issue #477, opened 2021-04-30
Flag for using an ordinary `dict` in the API response


brno32:
Currently, `_call_api` returns results as an `OrderedDict`. Could an optional flag be added to instruct the class to not make this transformation when pulling down the response as JSON?

sohithkumar:
@brno32 @jon-wobken  I have raised a pr which will make the json response to use dict if needed by the user. 
I created a object_pairs_hook parameter which by default uses OrderedDict (to be consistent with default json response by the package. User can pass either dict or None option to this parameter while calling if a dict json response is required.

jon-wobken:
This is resolved in >=1.12.0

PR #548, opened 2022-05-20
Add optional flag use ordinary dict

sohithkumar:
closes: #477 
Issue: Flag for using an ordinary dict in the API response #477 raised by @brno32 
Situation: json output is being decoded as OrderedDict with no option to override it. Users who need to use dict need an option to change the json output to use dict.
Solution: create a option object_pairs_hook parameter whose default value is OrderedDict. Users can change it to use dict or keep it None since the by default json result will use dict if object_pairs_hook is None.
@jon-wobken

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -8,9 +8,7 @@ import logging
 import re
 from collections import OrderedDict, namedtuple
 from urllib.parse import urljoin, urlparse
-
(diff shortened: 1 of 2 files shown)


brno32:
I want it!

jon-wobken:

jon-wobken (approved):

PR #563, opened 2022-07-19
Remove leftover reference to Python 3.5 support

edmorley:
Support for Python 3.5 was dropped in #556, so it should not be listed as a supported version in the README.


jon-wobken:
@edmorley  - Thanks for catching this. Will add in for the next release and probably take a finer look at the rest of the readme.

PR #594, opened 2022-11-12
Fix infinite refresh session issue

### LINKED ISSUE #569: infinite recursion when login returns 401
since version 1.12.0, simple-salesforce tries to refresh the token each time the login returns 401. 
Whenever the reason for getting a 401 is not token expiry, this generates an infinite recursion here: https://github.com/simple-salesforce/simple-salesforce/blame/master/simple_salesforce/api.py#L584

hkajiwara:
Refresh session only when getting the status 401 and the error code INVALID_SESSION_ID
Fixes #569

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -189,6 +189,7 @@ class Salesforce:
                 SalesforceLogin,
                 session=self.session,
                 username=username,
+                password=password,
                 consumer_key=consumer_key,
                 consumer_secret=consumer_secret,
                 proxies=self.proxies,
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -9,7 +9,9 @@ import re
 from collections import OrderedDict, namedtuple
 from functools import partial
 from urllib.parse import urljoin, urlparse
+
 import requests
+
 from .bulk import SFBulkHandler
 from .exceptions import SalesforceGeneralError
 from .login import SalesforceLogin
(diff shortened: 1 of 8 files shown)


simple_salesforce/api.py
```diff
@@ -603,8 +603,10 @@ def _call_salesforce(self, method, url, name="", **kwargs):
 
         if self._salesforce_login_partial is not None \
                 and result.status_code == 401:
-            self._refresh_session()
-            return self._call_salesforce(method, url, name, **kwargs)
+            error_details = json.loads(result.text.lstrip('[').rstrip(']'))
+            if error_details['errorCode'] == 'INVALID_SESSION_ID':
```
ericbn, line 607: I know Salesforce can have two different types of response here, depending on which endpoint was used: returning a list or just one object directly. But are you sure the list will always only have one object? Otherwise removing the `[]` JSON array characters can yield an unparseable JSON.

hkajiwara: According the [developer guide](https://developer.salesforce.com/docs/atlas.en-us.240.0.api_rest.meta/api_rest/errorcodes.htm), the response seems to have the array chars.

smsearcy: Is there a reason not to decode the JSON list and get the first item, in case there are multiple items?

For example:
```python
error_list = json.loads(result.text)
error_details = error_list[0]  # this could be combined with either of the other statements
if error_details['errorCode'] == 'INVALID_SESSION_ID':
```

ericbn: This can be simplified to:
```python
            response_content = result.json()
            if response_content[0]['errorCode'] == 'INVALID_SESSION_ID':
```

jon-wobken (approved):

PR #597, opened 2022-11-16
Use f-strings

ericbn:
since minimal supported version is already Python 3.6. See https://docs.python.org/3.6/reference/lexical_analysis.html#formatted-string-literals

### DIFF
--- a/simple_salesforce/login.py
+++ b/simple_salesforce/login.py
@@ -69,8 +69,8 @@ def SalesforceLogin(
 
     if sf_version.startswith("v"):
         error_msg = (
-            "Invalid sf_version specified ({version}). Version should not "
-            "contain a leading 'v'".format(version=sf_version)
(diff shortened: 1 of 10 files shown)


simple_salesforce/login.py
```diff
@@ -69,51 +69,42 @@ def SalesforceLogin(
...
-        client_id = "{prefix}/{app_name}".format(
-            prefix=DEFAULT_CLIENT_ID_PREFIX,
-            app_name=client_id)
+        client_id = f'{DEFAULT_CLIENT_ID_PREFIX}/{client_id}'
     else:
         client_id = DEFAULT_CLIENT_ID_PREFIX
 
-    soap_url = soap_url.format(domain=domain,
-                               sf_version=sf_version)
-
```
ericbn, line 88: Moved soap_url down, close to where it's first used.

PR #599, opened 2022-11-24
Use pathlib.Path

ericbn:
also:
- default encoding in bytes.decode is already 'utf-8'. See https://docs.python.org/3.6/library/stdtypes.html#bytes.decode

Available at least since Python 3.6, which is the minimal supported version. See https://docs.python.org/3.6/library/pathlib.html#pathlib.Path.read_bytes


simple_salesforce/metadata.py
```diff
@@ -322,16 +321,11 @@ def _read_deploy_zip(zipfile):
...
+            raw = zipfile.read()
         else:
-            file = open(zipfile, 'rb')
-            should_close = True
-        raw = file.read()
-        if should_close:
-            file.close()
-        return b64encode(raw).decode("utf-8")
+            raw = Path(zipfile).read_bytes()
+        return b64encode(raw).decode()
```
ericbn, line 328: Simplified code to not need the should_close variable. Also, it's opening and closing the file in the Path.read_bytes method.

I preferred this simpler solution than:
```
        try:
            raw = file.read()
        finally:
            if should_close:
                file.close()
```
which is equivalent. (code before commit was missing the finally block)

simple_salesforce/tests/test_api.py
```diff
@@ -859,8 +890,9 @@ def test_md_deploy_failed_status_code(self, mock_read_zip):
         client = Salesforce(session_id=tests.SESSION_ID,
                             instance=tests.INSTANCE_URL,
                             session=session)
+        zip_path = Path(__file__).parent / 'sample.zip'
         with self.assertRaises(Exception):
-            client.deploy("path/to/fake/zip.zip", sandbox=False)
+            client.deploy(str(zip_path), sandbox=False)
```
ericbn, line 895: Changes tests above to use an actual minimal zip file, instead of mocking _read_deploy_zip, and added another test passing an opened file handler instead of a path. This adds coverage to the _read_deploy_zip method.

ericbn:
Will update the PR match the code from the release/v1.12.3 branch.

simple_salesforce/login.py
```diff
@@ -195,8 +196,7 @@ def SalesforceLogin(
                 )
             }
         if privatekey_file is not None:
-            with open(privatekey_file, 'rb') as key_file:
-                key = key_file.read()
+            key = Path(privatekey_file).read_bytes()
```
jon-wobken, line 199: I am not as familiar with pathlib but only concern is if this would prevent the use of s3 or other cloud storage buckets

jon-wobken: There isn't any code re: remote file paths in this package. It was just one of the main things I saw when quickly looking into the differences between the two but digging a bit deeper it is not applicable here as it would be handled outside of this package.

PR #616, opened 2023-01-30
Support to resources exposed as text/csv

haroldo-bonette:
Some Salesforce resources like sobjects/EventLogFile/[id]/LogFile are exposed as base64 (csv) format.
This change decodes the byte format into string and converts it to json, the json string is then parsed using json.loads which returns an OrderedDict

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ dist/
 .coverage
 .tox/
 tags
+.env
 .env/
 .tags
 .tags_sorted_by_file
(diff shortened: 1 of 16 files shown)


jon-wobken:
Pandas is such a large library I am hesitant to force it as a requirement and if this is a needed functionality either handle it they a different means or using a smaller package.

haroldo-bonette:
> Pandas is such a large library I am hesitant to force it as a requirement and if this is a needed functionality either handle it they a different means or using a smaller package.

@jon-wobken I have refactored it and removed Pandas

PR #677, opened 2023-08-21
Updates to make bulk v2 upserts actually function

siarizar:
Main updates:

- Use UTF-8 encoding
- Removing chunking (messes CSV formatting)


gunnypatel:
@siarizar was there a reason you closed this PR? The changes to use `datetime` and encoding data to UTF-8 were exactly what I was looking for.

PR #705, opened 2024-01-12
Include "fields" as a query param in GET requests

framarzar:
Salesforce REST API allows us to specify the fields of a record we want to retrieve.

Example extracted from their [documentation](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_get_field_values.htm):

`curl https://MyDomainName.my.salesforce.com/services/data/v59.0/sobjects/Account/001D000000INjVe​?fields=AccountNumber,BillingPostalCode -H "Authorization: Bearer token"`

I've added a new "fields" argument on both methods from SFType: `get() & get_by_custom_id()`. So, if we want to retrieve data of certainn  this allows us to retrieve the data of that fields. For instance:

`salesforce_instance.Account.get('001D000000INjVe', fields="AccountNumber,BillingPostalCode")`

This is an optional argument, so if fields=None, the response will include all the fields of that SFType.

### DIFF
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -1026,10 +1026,8 @@ class SFType:
         * record_id -- the Id of the SObject to get
         * headers -- a dict with additional request headers.
         """
-        record_url = urljoin(self.base_url, record_id)
-
         result = self._call_salesforce(
-            method='GET', url=record_url,
+            method='GET', url=urljoin(self.base_url, record_id),
             headers=headers, **kwargs
             )
         return self.parse_result_to_json(result)
@@ -1050,7 +1048,7 @@ class SFType:
                              as an External ID
         * custom_id - the External ID value of the SObject to get
         * headers -- a dict with additional request headers.
-         """
+        """
         custom_url = urljoin(self.base_url,
                              f'{custom_id_field}/{custom_id}'
                              )
--- a/simple_salesforce/api.py
+++ b/simple_salesforce/api.py
@@ -800,20 +800,26 @@ class SFType:
             )
         return self.parse_result_to_json(result)
 
-    def get(self, record_id, headers=None):
+    def get(self, record_id, headers=None, fields=None):
         """Returns the result of a GET to `.../{object_name}/{record_id}` as a
         dict decoded from the JSON payload returned by Salesforce.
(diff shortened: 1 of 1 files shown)


simple_salesforce/api.py
```diff
@@ -800,20 +800,27 @@ def describe_layout(self, record_id, headers=None):
             )
         return self.parse_result_to_json(result)
 
-    def get(self, record_id, headers=None):
+    def get(self, record_id, headers=None, fields=None):
```
jon-wobken, line 803: I think we can actually just refactor this to pass **kwargs in def get as well as in the result = self._call_salesforce and pass **kwargs there as well. Then the user can include any potential additional parameters in the future and pass fields such that params={'fields':'field1,field2,field3,etc'}

simple_salesforce/api.py
```diff
@@ -800,20 +800,27 @@ def describe_layout(self, record_id, headers=None):
...
+
         result = self._call_salesforce(
-            method='GET', url=urljoin(self.base_url, record_id),
+            method='GET', url=record_url,
             headers=headers
             )
         return self.parse_result_to_json(result)
 
-    def get_by_custom_id(self, custom_id_field, custom_id, headers=None):
+    def get_by_custom_id(self, custom_id_field, custom_id, headers=None,
```
jon-wobken, line 822: I think we can actually just refactor this to pass **kwargs in def get_by_custom_id as well as in the result = self._call_salesforce and pass **kwargs there as well. Then the user can include any potential additional parameters in the future and pass fields such that params={'fields':'field1,field2,field3,etc'}

jon-wobken (changes requested):
I think we can actually just refactor this to pass **kwargs in both functions as well as in the result = self._call_salesforce and pass **kwargs there as well. Then the user can include any potential additional parameters in the future and pass fields such that params={'fields':'field1,field2,field3,etc'}

simple_salesforce/api.py
jon-wobken: dug into this a bit more. I think the better option is to just pass **kwargs therefore users can just input as so

sf.Account.get('recordid',params='fields=Name,''RecordTypeId')

Issue #650, opened 2023-04-27
Question: Is there an easy way to get a field of an object?


men6288:
Hi folks,

I am currently using the describe function to get fields related to an object. Is there a more direct way to pull a field and all the picklist values?

Ex of what I am using:
`sf.Account.describe()['fields']`

I then loop through these fields to get the picklist values.

PR #712, opened 2024-02-17
Issue: #650  Created a function that return the all the picklist values when passed with the describe() result and the field api name.

ZadeThakur:
To get a picklist values of a field, the user have to loop through all the fields and then extract out the picklist values. 

This function takes care of that process. The user needs to pass the describe result and the field api name(they want the values of) and the function will return a dictionary with the field name as the key and a list of picklist value associated with that field.

It returns a empty dictionary if the field is not found and it the field is not a picklist field.


ZadeThakur:
@jon-wobken Does this solves the issue?

PR #722, opened 2024-03-29
correct query example in readme

AlexisFavre:


wyaeld:
@jon-wobken can you please merge this single doc with, without it you get something like

```
Malformed request https://xxx.my.salesforce.com/services/data/v59.0/jobs/query. Response content: [{'errorCode': 'API_ERROR', 'message': 'Value does not match expected type at [line:1, column:22]'}]
```

Which is nearly impossible to debug, and there are few examples elsewhere of correct bulk2 usage.

Issue #684, opened 2023-09-16
bulk2 delete function not working with records parameter


michaelhollander:
To recreate:
```
records = [{"Id": "0062300000IdMo2AAF"},{"Id": "0062300000IdMo3AAF"}]

sf.bulk2.Contact.delete(records = records)

```
What I expect: the deletion to work as normal.

What actually happens: I get an error on line 646 of bulk2. The problem seems to be that although bulk2 allows you to pass in the records to delete using the records parameter, that param isn't actually used on [on line 646](https://github.com/simple-salesforce/simple-salesforce/blob/master/simple_salesforce/bulk2.py#L645C8-L646C63)

```
if operation in (Operation.delete, Operation.hard_delete):
            with open(csv_file, "r", encoding="utf-8") as bis:

```
The other operations don't appear to have this problem.

Issue #735, opened 2024-06-17
Bulk2 Delete: not possible to use records parameter instead of csv_files


sauerchextern:
Unfortunately it is not possible to use the records parameter for bulk2 delete. It can only be used with csv_file.

Here is the reason why:
https://github.com/simple-salesforce/simple-salesforce/blob/1d28fa18438d3840140900d4c00799798bad57b8/simple_salesforce/bulk2.py#L871

zhiweio:
> Unfortunately it is not possible to use the records parameter for bulk2 delete. It can only be used with csv_file.
> 
> Here is the reason why:
> 
> https://github.com/simple-salesforce/simple-salesforce/blob/1d28fa18438d3840140900d4c00799798bad57b8/simple_salesforce/bulk2.py#L871

Hi sauerchextern, Salesforce Bulk API V2 is designed for exchanging data through CSV files, You can find more details about it here: [Salesforce Bulk API V2 Documentation](https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/datafiles_prepare_data.htm). For a large number of records, it’s best to convert them into CSV files and use the bulk API. However, for just a few records, the REST API would be a better choice.

zhiweio:
> Unfortunately it is not possible to use the records parameter for bulk2 delete. It can only be used with csv_file.
> 
> Here is the reason why:
> https://github.com/simple-salesforce/simple-salesforce/blob/1d28fa18438d3840140900d4c00799798bad57b8/simple_salesforce/bulk2.py#L871

That's a thoughtless implementation. Thanks for reporting it. I'm happy to fix it this weekend.

PR #736, opened 2024-06-22
Refine CSV files splitting implementation in 'bulk2' 

zhiweio:
The previous CSV splitting method was not careful and couldn't properly handle CSV files and data that included special characters. 
I've now completely improved it using the standard 'csv' library for reading and writing. This update allows for various column delimiters and line endings. 
Additionally, I fixed the bugs mentioned in issues #684 #735

@jon-wobken Could you please review it, thanks a lot


zhiweio:
Fix #684

PR #750, opened 2024-10-07
Update README.rst

Abdur-rahmaanJ:
Clarify how to get security token


h0h0h0 (approved):
Had to do the same thing recently and this clarification helps a ton.

Abdur-rahmaanJ:
Thanks for the testimonial! ^^

williambritt:
It might be valuable to include a link to the Salesforce documentation as well:
https://help.salesforce.com/s/articleView?id=xcloud.user_security_token.htm&type=5

Additionally, I cannot seem to find documentation to support this. But I want to say, the token is only re-generated on password reset if you have already requested a security token.

h0h0h0:
-Thanks!

PR #771, opened 2025-04-02
Fix restful method not to error out if status code is 204

gauravrmazra:
When any of the restful api returns 204, this was erroring out with below error:
"""
Expecting value: line 1 column 1 (char 0)
"""
The reason was trying to parse empty string as json.


gauravrmazra:
@jon-wobken Tagging you since you have been maintaining this library lately.

PR #778, opened 2025-06-09
Visualization of the simple-salesforce codebase

ivanmilevtues:
In this PR I have added high-level diagrams representing the simple-salesforce codebase.
You can see how they render in the github markdown here: https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/simple-salesforce/on_boarding.md

The idea behing these diagrams is to help people get up-to-speed with the existing codebase. Me and a friend are wokring on this by leveraging Static Analysis and LLMs. We believe that visuals are the best way to get started thus the diagrams driven documentaiton, further we think that having the big picture is very important that is why the first diagram is very abstract and then a person can dive deeper into each component.

Any feedback is more than welcome! I'd usually start with a disucssion, but they don't seem to be enabled for this repository that is why I decided to go ahead and create a PR.

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people.

### DIFF
--- a/.codeboarding/Salesforce Authentication & Session Management.md
+++ b/.codeboarding/Salesforce Authentication & Session Management.md
@@ -1,0 +1,77 @@
+```mermaid
+graph LR
+    Authentication_Orchestrator["Authentication Orchestrator"]
+    SOAP_Login_Handler["SOAP Login Handler"]
+    Token_Login_Handler["Token Login Handler"]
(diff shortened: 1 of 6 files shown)


ivanmilevtues:
Just a quick update, we are open-source now :)
https://github.com/Shopify/shopify_python_api

PR #784, opened 2025-08-10
Add modular DML functionality

tim-kornish-cloud:
I have created two functions in the bulk.py file that modularize the usage of the DML (Data Manipulation Language) operations: insert/upsert/update/delete.

I have not deleted or altered any existing functionality with the exception of an issue in a unit test in test_bulk.py and some lines have been slightly reformatted by my text editor adding white space or a new line.

This new functionality allows the removal of a hardcoded insert/upsert/update/delete from an execution script with all the pre and post processing that a developer can include with it, and placing that functionality into a custom function to perform the same reporting across every object for these 4 operations.

I have added to the test_bulk .py 4 unit test and fixed a small error where the update test has operation = 'upsert' instead of = 'update'

Since there is no hard_delete unit tests in the test_bulk.py I left those out. I'm not sure how to run the unit test in the test folder to check my included unit test.
I have only tested and confirmed the functionality for the 4 operations listed with mock data from mockaroo.com, inserting/updating/upserting/deleting accounts and contacts into a dev org. 

Before passing or rejecting these additions, please let me know if there any explanations or additions someone would like to see.

### DIFF
--- a/simple_salesforce/bulk.py
+++ b/simple_salesforce/bulk.py
@@ -85,7 +85,7 @@ class SFBulkHandler:
 
             The main purpose of this function is to
             build customizable reporting functions and reduce code reuse
-            while reducing code in indivual execution scripts mainly with pandas
+            in indivual execution scripts mainly with pandas
 
         Arguments:
 
--- a/simple_salesforce/bulk.py
+++ b/simple_salesforce/bulk.py
@@ -69,11 +69,11 @@ class SFBulkHandler:
                    object_name: str,
                    dml: str,
                    data: BulkDataAny,
+                   external_id_field: str = None,
                    batch_size: int = 10000,
                    use_serial: bool = False,
                    bypass_results: bool = False,
-                   include_detailed_results: bool = False,
-                   external_id_field: str = None
+                   include_detailed_results: bool = False
                    ):
         """ Perform any DML operation on any custom or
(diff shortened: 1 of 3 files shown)


tim-kornish-cloud:
I also updated the readme to include a pseudo code example of what using this new submit_dml funciton could look like along with real examples for each of the four main DML operations. I can upload more example python how I process and split the data for processing to the readme in the future

simple_salesforce/bulk.py
```diff
@@ -655,3 +699,30 @@ def query_all(
...
+        """ modular bulk dml operations -
+            perform insert/upsert/update/delete
+            on any standard and custom objects in Salesforce."""
+        if function_name == 'upsert' and external_id_field != None:
+            return getattr(self, function_name)(data,
+                                                batch_size,
+                                                use_serial,
+                                                bypass_results,
+                                                include_detailed_results,
+                                                external_id_field)
```
tim-kornish-cloud, line 722: The external_id_field needs to come after the parameter data and before batch_size, I have this fixed on my personal computer by forgot to copy it over, this function should still work for insert/update/delete but will fail on upsert due to the misordering of parameters when in this if statement

tim-kornish-cloud:
closing, to re-upload pull request with the external_id_patch

PR #807, opened 2025-10-31
add retry/timeout support and default session builder

jkasireddy:
- Add optional kwargs: request_timeout, max_retries, backoff_factor, status_forcelist
- Build a requests.Session with urllib3 Retry when no custom session is supplied
- Inject default timeout in call_salesforce when per-call timeout not passed
- Backward compatible; docs/tests included"


jkasireddy (commented):
Looks Good to me

PR #731, opened 2024-06-04
add ability to query listview

h0h0h0:
Enables querying listviews via simple-salesforce #730

PR #811, opened 2025-12-09
Listview support 

h0h0h0:
Added some docs and tests along with listview support on the custom SF object. 

Simple-salesforce did not have a great way to get direct access to the listview REST methods that were there so I wanted to add them.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -177,6 +177,36 @@ Make sure to have all the required fields for any entry. The `Salesforce API`_ h
 .. _Salesforce HTTP Status Code: http://www.salesforce.com/us/developer/docs/api_rest/Content/errorcodes.htm
 .. _Salesforce API: https://www.salesforce.com/developer/docs/api/
 
+Record Management - Listviews
+--------------------------
+
+
+To view available listviews for a given salesforce object use  
+
+.. code-block:: python
+
+    sf.Contact.listviews({'LastName':'Smith','Email':'example@example.com'})
+
+This will return a dictionary such as ``{'done': 'true', 'listviews':[{'id': 'someid', 'url': 'etc'}], 'success': True}``
+
+To get some information for a given listview, you can get basic info using `listview_basicinfo` and for more detailed info use `listview_detailed`:
+.. code-block:: python
+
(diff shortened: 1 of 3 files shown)


h0h0h0:
@nickcatal this is a replacement for #731 

I added docs and tests as well as properly situating the methods on the SFObj.

ramonvanschaik:
Hey. First of all, thanks for this PR. As even though it's not merged, it gave me some handles to at least get the ListView data with some scripting.

Got one remark: for normal Object query results, I'm usually able to parse the results to a Pandas DataFrame by doing `df = pd.DataFrame(data['records']).drop('attributes', axis=1)` as per documentation. For the result I got from your code, I had to do a bit more list magic in order to get .

```
# result is quite deeply nested. Use this to get something that can be put in a dataframe
r = list(map(lambda x: x["columns"], data["records"]))
r = list(map(lambda x: {e["fieldNameOrPath"]: e["value"] for e in x}, r))
df = pd.DataFrame(r)
```

Definitely not blocking for the PR, but I suppose many people would want to use this functionality catch the result to a dataframe.

h0h0h0:
@ramonvanschaik Thanks for the feedback there. Search results from the REST API don't have as much decoration and supportive constructs as the listview results have. I had a chuckle because I ran into the same exact foot gun you pointed out above when I implemented this long ago. 

What do you and @nickcatal think about this approach?

I'll add a required `return_style` parameter to the listview calls. This would create a small speedbump decision to be made. You could select a `standard`-style which returns the same conventional `.query` return construct. This 'standard' style would have the library translate the data itself.  Or a `detailed`-style which would have the library return the listview-Rest style without any further translation.

ramonvanschaik:
Including it in the library sounds like a good way forward! Thanks for that. 

To accommodate notebook analysts with less python knowledge, I would add some snippet to the documentation like "run this to get your listview into a dataframe". I think there's already a pandas section in the readme.

h0h0h0:
@nickcatal I know you're really busy. Just wanted to float this up to you. Let me know if you'd like a hand with some other stuff.

PR #812, opened 2025-12-14
feat: Add Support for Salesforce CLI Authentication (MFA/SSO Friendly)

mounishreddy27:
## Motivation
Salesforce is actively deprecating the legacy SOAP login API (username/password). Users are now seeing warnings such as:
> *"SOAP API login() is disabled by default... and will be retired in the Summer '27 release."*

Combined with strict MFA (Multi-Factor Authentication) enforcement, this makes the traditional login method difficult or impossible for many users. This PR provides a future-proof alternative by allowing `simple-salesforce` to use the authenticated session from the **Salesforce CLI**, which handles MFA and SSO natively.

## The Solution
This PR allows `simple-salesforce` to piggyback off the official **Salesforce CLI** (`sf` or `sfdx`). If a user is already logged in locally (via web browser flow), this library can now utilize that valid session instantly.

## Changes
* **`api.py`**: Updated `Salesforce.__init__` to accept `use_cli=True` and `target_org='alias'`.
* **`util.py`**: Added `get_cli_session()` to securely fetch the access token and instance URL via a subprocess call.
* **Tests**: Added full test coverage for success scenarios, aliases, missing CLI tools, and runtime errors.

## Usage Example
```python
from simple_salesforce import Salesforce

# No password or token needed if you are logged in via CLI
sf = Salesforce(use_cli=True)

# Works with specific aliases
sf = Salesforce(use_cli=True, target_org='MySandbox')

### DIFF
--- a/simple_salesforce/tests/test_util.py
+++ b/simple_salesforce/tests/test_util.py
@@ -149,13 +149,13 @@ class TestUtilCli(unittest.TestCase):
 
     @patch('shutil.which')
     def test_cli_not_found(self, mock_which):
-        """Test that we raise an Exception if neither sf nor sfdx is found"""
+        """Test that we raise ValueError if neither sf nor sfdx is found"""
         
         # Setup: Return None for both checks
         mock_which.return_value = None
         
-        # Verify it raises the generic Exception you defined
-        with self.assertRaises(Exception) as context:
+        # Verify it raises ValueError
+        with self.assertRaises(ValueError) as context:
             get_cli_session()
         
         self.assertIn("Salesforce CLI not found", str(context.exception))
--- a/simple_salesforce/tests/test_util.py
+++ b/simple_salesforce/tests/test_util.py
@@ -11,8 +11,12 @@ from simple_salesforce.exceptions import (SalesforceExpiredSession,
                                           SalesforceRefusedRequest,
                                           SalesforceResourceNotFound)
 from simple_salesforce.util import (date_to_iso8601, exception_handler,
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: simple_salesforce/tests/test_util.py [hunk 1]
```diff
@@ -1,7 +1,9 @@
 """Tests for simple-salesforce utility functions"""
 import datetime
+import json
+import subprocess
```
h0h0h0, line 4: Thank you - i always got bothered by the order of the imports.

andy-pi:
Thanks for the work on this. I would like to be able to continue to use simple_salesforce heedlessly, for my use case of, I don't have Salesforce CLI or access to a browser  - so I'm a little confused about if this PR will solve the issue? @mounishreddy27 any advice?

mounishreddy27:
> Thanks for the work on this. I would like to be able to continue to use simple_salesforce heedlessly, for my use case of, I don't have Salesforce CLI or access to a browser - so I'm a little confused about if this PR will solve the issue? @mounishreddy27 any advice?

@andy-pi 
Based on your description, I don’t think this PR directly addresses your use case.

This PR focused on enabling Salesforce CLI authentication for MFA/SSO scenarios, whereas your workflow depends on headless/non-CLI authentication without browser access. Those are somewhat separate concerns.

So unless you plan to adopt Salesforce CLI–based auth, this PR likely doesn’t provide much benefit for your current setup.

andy-pi:
thanks  @mounishreddy27 thats clear!

PR #816, opened 2026-02-03
fixed the authentication issue related to client credentials via exte…

beau32:
The current simple-salesforce implementation only supports username/password authentication flows, which is incompatible with Salesforce's newer External Connected App configuration that uses OAuth 2.0 Client Credentials flow for server-to-server authentication without requiring user credentials.

This PR adds support for the OAuth 2.0 Client Credentials flow, avoid username and password exposure during the authentication process using only:
Client ID (Consumer Key)
Client Secret (Consumer Secret)

### DIFF
--- a/simple_salesforce/login.py
+++ b/simple_salesforce/login.py
@@ -3,7 +3,7 @@
 Heavily Modified from RestForce 1.0.0
 """
 from typing import Any, Dict, Optional, Tuple, Union, cast
-import base64
+import base64,logging
 
 DEFAULT_CLIENT_ID_PREFIX = 'simple-salesforce'
 
@@ -21,6 +21,8 @@ from .api import DEFAULT_API_VERSION
 from .exceptions import SalesforceAuthenticationFailed
 from .util import Headers, Proxies, getUniqueElementValueFromXmlString
 
+logging.basicConfig
+logger = logging.getLogger(__name__)
 
 # pylint: disable=invalid-name,too-many-arguments,too-many-locals,too-many-branches
 def SalesforceLogin(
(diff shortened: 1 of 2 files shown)


simple_salesforce/login.py
```diff
@@ -229,6 +244,16 @@ def SalesforceLogin(
...
+
+        logger.debug('token data: %s',
+                    {'grant_type': 'client_credentials',
+                      'client_id': consumer_key,
+                      'client_secret': consumer_secret,
+                      'username': username,
+                      'password': password,
+                      'instance_url': instance_url,
+                        'domain': domain})
+
```
arbitraryconstants, line 256: Looks like you forgot to remove some debug logging?

beau32: this will only show when logger is set to debug mode

arbitraryconstants: Personally, I wouldn't want to log credentials, even in debug mode.

arbitraryconstants: imo this introduces a security vulnerability into the package.

mayday6971:
This PR is pretty much unnecessary. You don't need to pass a username or password when you are using a client credentials configuration on the External Client App. If you configure your External Client App and check the box for "Enable Client Credentials Flow" and set the Username that the External App should run as.

image

Note that Salesforce wants this method to be used for server to server communication. 
https://help.salesforce.com/s/articleView?id=xcloud.remoteaccess_oauth_client_credentials_flow.htm&type=5

Docs on Client Credential Flow:
https://help.salesforce.com/s/articleView?id=xcloud.configure_client_credentials_flow_for_external_client_apps.htm&type=5

Code showing this working in the current version.

`
sf = Salesforce(
  consumer_key=sf_consumer_key,
  consumer_secret=sf_consumer_secret,
  client_id='client-name',
  domain=sf_domain,
  version='66.0'
)
`

